PR #2, opened 2020-10-02
Updates from the plugin generator

agrare:
Ran `rails generate manageiq:provider ManageIQ::Providers::Autosde --path ../ --vcr --manager-type=storage`


agrare:
cc @gregoryb-at-ibm

PR #3, opened 2020-10-02
Add typhoeus gem depdendency

agrare:
Fixes https://github.com/Autosde/manageiq-providers-autosde/issues/1


agrare:
cc @gregoryb-at-ibm

PR #20, opened 2020-10-02
Fixes #6 removed unused event_catcher

### LINKED ISSUE #6: Remove boilerplate event_catcher
The EventCatcher isn't implemented and is just what the provider generator scaffolding builds, it should be removed so a worker doesn't run it

gregoryb-at-ibm:


agrare:
Should also remove `app/models/manageiq/providers/autosde/storage_manager/event_catcher`

PR #22, opened 2020-10-03
Fixes #17 delete unneeded file

### LINKED ISSUE #17: Cleanup manageiq-providers-autosde.rake task
```
namespace "manageiq:providers:autosde" do
  desc "bk Explaining what the task does"
  task :my do
    puts "AAAAAAAAAAAAAAAA"
    puts File.dirname(__FILE__)
    puts ManageIQ::Providers::Autosde::Engine.root
  end
end
```

borisko123:


gregoryb-at-ibm:
Can we also get rid of taks/manageiq-providers-autosde.rake?

PR #23, opened 2020-10-05
Remove unused event_catcher runner/stream

agrare:
These classes are just the templates generated by the provider generator and aren't implemented yet


agrare:
cc @gregoryb-at-ibm

PR #24, opened 2020-10-05
Move openapi_client out of app/models

agrare:
This shouldn't live in app/models/ since it is a library, should be in lib/

Ideally we should rename the `OpenapiClient` classes to something else since this is the default name for the openapi generator and might collide


agrare:
cc @gregoryb-at-ibm 

This passes specs locally

PR #25, opened 2020-10-05
Added empty asset_detail to physical_storage inventory collection

gregoryb-at-ibm:
We don't use the asset detail in our provider, but various existing function that deal with physical_storage expect it to be there.

### DIFF
--- a/app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
+++ b/app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
@@ -27,10 +27,13 @@ class ManageIQ::Providers::Autosde::Inventory::Parser::StorageManager < ManageIQ
   def physical_storages
     collector.physical_storages.each do |physical_storage_hash|
       system_type_uuid = physical_storage_hash.delete(:system_type_uuid)
-      persister.physical_storages.build(
+      physical_storage = persister.physical_storages.build(
         :physical_storage_family => persister.physical_storage_families.lazy_find(system_type_uuid),
         **physical_storage_hash
       )
+
+      # asset detail is required for physical_storage, even if we don't use it.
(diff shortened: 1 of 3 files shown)


gregoryb-at-ibm:
@agrare , could you PTAL to see you approve this solution?

agrare:
@gregoryb-at-ibm what functionality expects asset details to exist?  Wonder if we should be changing that rather than stubbing it out here

llivne:
> @gregoryb-at-ibm what functionality expects asset details to exist? Wonder if we should be changing that rather than stubbing it out here

@agrare if you look on manageiq-ui-classic/app/helpers/physical_storage_helper/textual_summary.rb you will see that it's tightly coupled with asset details. we thought about a different solution, individual textual summary (see https://github.com/ManageIQ/manageiq-ui-classic/pull/7325) but it requires redesign and thinking so using the existing form looks for me like the right solution for the short term. I did opened an internal ticket for the team to design how and if we want to implement the individual textual summary, and obviously it will be with miq collaboration.

agrare:
https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/helpers/physical_storage_helper/textual_summary.rb#L45-L47 being an example

If this is just so that will return `nil` instead of raise an exception I think it'd be better to update the UI to handle a nil `asset_detail` e.g.: `@record.asset_detail&.product_name`   @skateman @himdel thoughts?

skateman:
I guess I'm okay with that

llivne:
@gregoryb-at-ibm please close this issue, I opened https://github.com/ManageIQ/manageiq-ui-classic/pull/7405 to solve this issue as discussed in here

gregoryb-at-ibm:
Fixed in ManageIQ/manageiq-ui-classic#7405

PR #27, opened 2020-10-06
Fix ems refresh following commit 387abe2

llivne:
In commit https://github.com/ManageIQ/manageiq/commit/387abe2360 array of queue names was delegate to parent_manager. Our manager doesn't have a parent-manager and so we couldn't create our manager.


gregoryb-at-ibm (approved):

gregoryb-at-ibm:
@agrare PTAL

agrare (approved):

PR #31, opened 2020-10-07
Fixes issue openapi generator output location

borisko123:
The openapi gem was moved to lib
This gem is generated by task openapi_oas.rake
This task should be fixed apropriately.
Also, make gemfile be familiar with this gem


gregoryb-at-ibm:
This will also make it easier to transition when we put our generated code in a gem on rubygems.org.

Gemfile
```diff
@@ -1,4 +1,6 @@
 # Declare your gem's dependencies in manageiq-providers-autosde.gemspec.
+#
```
gregoryb-at-ibm, line 2: We don't need an empty comment line.

Gemfile
```diff
@@ -10,3 +12,6 @@ gemspec
 
 # Load Gemfile with dependencies from manageiq
 eval_gemfile(File.expand_path("spec/manageiq/Gemfile", __dir__))
+
+# load openapi gem
+gem 'openapi_client', :path => 'lib/autosde_oas_client/generated'
```
gregoryb-at-ibm, line 17: openapi_client is too general. Many application use openapi.
 Please name this soemthing like autosde_openapi_client.

gregoryb-at-ibm (changes requested):

Gemfile
agrare: While this is in the same repository I think we should leave it as a "normal" library rather than adding a path based gem.
We shouldn't be too far away from extracting this to a proper gem so I'd rather just wait for that.

@gregoryb-at-ibm is right though we should update the `.openapi-config.json` to take a `gemName` so it is something other than the default.

lib/tasks_private/autosde_oas.rake
```diff
@@ -7,7 +7,7 @@ namespace :autosde do
 
       root = ManageIQ::Providers::Autosde::Engine.root # ManageIQ::Environment::APP_ROOT
       # parent folder for generated stuff ( contains input spec file)
-      default_folder = "app/models/manageiq/providers/autosde/storage_manager/openapi_client"
+      default_folder = "lib/autosde_oas_client"
```
agrare, line 10: I forgot to update this when I moved the client out of `app/models` that was my bad

agrare:
Hey @borisko123 I think the travis failures are legit, what looks like is happening is you gave it a target directory of `lib/autosde_oas_client/generated/lib/autosde_openapi_client` but didn't change the target gemName so it is still called `OpenapiClient`.

`lib/autosde_oas_client/generated/lib/autosde_openapi_client/openapi_client/api_client.rb`
has:
```
module OpenapiClient
  class ApiClient
```

borisko123:
All **failed** checks by codeclimate are for generated code for OpenApi client
Actually, it should not be checked here.
When it will be separate gem, there will be no checks.
We may ignore these fails for now, IMHO

Gemfile
borisko123: I've generated openclient anew with name  'autosde_openapi_client'.

agrare:
@borisko123 yes we always ignore codeclimate/rubocop for generated code

PR #32, opened 2020-10-11
Added empty cloud_tenants and volume_availability_zones to Autosde EMS.

gregoryb-at-ibm:
Following the changes in https://github.com/ManageIQ/manageiq-ui-classic/commit/7123ca82a356cf60b358fe35e14220a810bcd009

In the mentioned commit, the API call sent by the form was changed. Instead of using `parent_manager.volume_availability_zones` it's now just `volume_availability_zones`. 
This causes an exception if the manager doesn't have volume_availability_zones. Before it would just return `nil` if there was no parent-manager.
Our EMS has no volume_availability_zones, and so we are getting an exception on the new-volume form.

In this PR I try to solve this by providing dummies for the missing fields.

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager.rb
@@ -44,6 +44,17 @@ class ManageIQ::Providers::Autosde::StorageManager < ManageIQ::Providers::Storag
   virtual_column :total_resources, :type => :integer
   virtual_column :resources_info, :type => :json
 
+  attribute :volume_availability_zones, :boolean
+  attribute :cloud_tenants, :boolean
+
+  def cloud_tenants
+    nil
+  end
+
+  def volume_availability_zones
+    nil
+  end
+
   class << model_name
     define_method(:route_key) { "ems_block_storages" }
     define_method(:singular_route_key) { "ems_block_storage" }


gregoryb-at-ibm:
@agrare PTAL and see if the solution approach is acceptable. Or maybe there's a better solution?

llivne:
> @agrare PTAL and see if the solution approach is acceptable. Or maybe there's a better approach?

@gregoryb-at-ibm @agrare, I think we should do this inside ManageIQ::Providers::StorageManager::BlockMixin (miq repo) and not only for autosde provider. cloud volume is a feature that all block storages have, so we need to handle this issue for all block storages. what do you think ??

agrare:
Oh good catch @gregoryb-at-ibm 

So I don't think we should do this in the `BlockMixin` because this is an issue for standalone storage managers, not all block managers.

Long term the core `ManageIQ::Providers::StorageManager` and `ManageIQ::Providers::NetworkManager` should be standalone by default, rather than belonging to a parent manager by default which is how they are right now.

For now you can reuse the associations from the CloudManager (has_many :cloud_tenants, has_many :volume_availability_zones, :class_name => "AvailabilityZone") and I'll open an issue to track refactoring the base storage and network managers.

agrare:
https://github.com/ManageIQ/manageiq/issues/20687

gregoryb-at-ibm:
@agrare 
Like so? https://github.com/ManageIQ/manageiq/pull/20698

agrare:
:+1: yes like that

gregoryb-at-ibm:
Doing the change in ManageIQ/manageiq#20698 instead.

PR #34, opened 2020-10-15
Use manageiq-style

bdunne:
Depends On: ManageIQ/miq_bot#506

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,10 +1,9 @@
----
 version: '2'
 prepare:
   fetch:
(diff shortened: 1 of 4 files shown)


agrare:
@gregoryb-at-ibm a PR description with `Depends on:` means the PR shouldn't be merged until the stated dependency has been merged.

@bdunne will this fail in the meantime?  I see the rubocop_base exists: https://raw.githubusercontent.com/ManageIQ/manageiq-style/master/.rubocop_base.yml 

If it will fail I can put in a revert PR

bdunne:
Yes, a couple of cops were renamed, so I would expect that to cause problems.

PR #37, opened 2020-10-21
[WIP]changed sizes to gb

IlyaKessel:
Added size conversion to GB


chessbyte:
@agrare does this align with the other providers?

agrare:
Unless otherwise noted all storage sizes in the db should be in bytes

Fryguy:
Also the rubocops should be addressed.

gregoryb-at-ibm:
So we need to redo this and normalize everything to bytes.

erantzabari:
We agree that it should all be presented in Bytes.

PR #38, opened 2020-10-21
Implementing remove physical storage

llivne:
Adding functionality for deleting supported physical storages via rest-API

links
-----------
the following are related PRs
https://github.com/ManageIQ/manageiq/pull/20713
https://github.com/ManageIQ/manageiq-api/pull/932
https://github.com/ManageIQ/manageiq-ui-classic/pull/7433

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
@@ -1,5 +1,12 @@
 class ManageIQ::Providers::Autosde::StorageManager::PhysicalStorage < ::PhysicalStorage
   supports :create
+  supports :delete
+
+  def raw_delete_physical_storage
+    ems = ext_management_system
+    ems.autosde_client.StorageSystemApi.storage_systems_pk_delete(ems_ref)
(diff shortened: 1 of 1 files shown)


gregoryb-at-ibm:
Looks fine to me.

gregoryb-at-ibm (approved):

app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
```diff
@@ -1,5 +1,12 @@
 class ManageIQ::Providers::Autosde::StorageManager::PhysicalStorage < ::PhysicalStorage
   supports :create
+  supports :delete
```
agrare, line 3: @llivne and here you can add your check that it is connected to an EMS:
```
supports :delete do
  unsupported_reason_add(:delete, _("The Physical Storage is not connected to an active Manager"))
end
```

agrare: We don't disconnect physical storages when an EMS is deleted though so unless someone manually created the record or deleted the EMS without calling destroy this can't happen (a physical storage without an EMS).

PR #48, opened 2020-11-17
support for newer version of autosde provider API

llivne:
We added additional API functionality to autosde that need to be reflected in miq autosde provider also.
the new functionality includes:
-job
-snapshot
-storageHost


gregoryb-at-ibm (approved):
@borisko123, we need to see about moving this to a separate gem.

PR #49, opened 2020-12-02
Implement Host Initiators

llivne:
Host Initiators (AKA physical storage hosts) are virtual entities that only exists as a set of definitions within the `PhysicalStorage' and hold the information about consumers/hosts that use (defined) the physical storage.

This PR includes the work for retrieving and presenting Host Initiators for a specific physical storage in manageiq.

g
h

Example data
---------
Below are diagrams that shows how we'd represent rather standard scenario. 

All items in white are already part of miq:
We have a single IRL storage-system Called SVC Storage 1.
It has 2 storage pools: Pool1 and Pool2.
Pool1 has two volumes: Volume1 and Volume2.

All items in blue are changes done in this PR:
The Physical Storage has a single host-initiator defined in it. It has been defined with two addresses: iqn-addr-1, and wwpn-addr-1. For the sake of brevity I won't use real IQN and WWPN addresses as they are too long. 

All items in peach are future drafts pending changes (not part of this PR):
The host-initiator in the storage-system has been mapped to Volume1.
A DB server consumes Volume1.
One of the FC addresses of the DB server is wwpn-addr-2. Remeber that this address was defined on the host-initiator above. This, together with the fact that the host-initiator was mapped to Volume1 allows the DB server to consume it.

========== using STI for SAN-Addresses ==========
In this diagram, the addresses are implemented as an STI.
There's a general SAN (Storage Area Network) Address table that holds all possible info for all address. It's expected to be a sparse table as concrete addresses will only populate their specific fields. 
So FcAddresses will have nil for their iscsi values such as IQN and CHAP. And the reverse holds true for IscsiAddresses.

links
----------
https://github.com/ManageIQ/manageiq-ui-classic/pull/7529
https://github.com/ManageIQ/manageiq/pull/20858
https://github.com/ManageIQ/manageiq-schema/pull/537


app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
```diff
@@ -26,6 +25,43 @@ def storage_resources
     end
   end
 
+  def addresses
+    @addresses ||= @manager.autosde_client.StorageHostApi.storage_hosts_get.map do |addresses|
+      addresses_array = []
```
gregoryb-at-ibm, line 30: `addresses_array`  here is redundant. Each address should go into its own hash. No need to group them into arrays.
Look how it's done in all other collectors.

app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
```diff
@@ -26,6 +25,43 @@ def storage_resources
     end
   end
 
+  def addresses
+    @addresses ||= @manager.autosde_client.StorageHostApi.storage_hosts_get.map do |addresses|
```
gregoryb-at-ibm, line 29: You are calling this API twice. Once here and once in physical_storage_consumers. For performance, it's better to only call it once, and then process the results. 
So I suggest you unify the methods `addresses` and `physical_storage_consumers`.

app/models/manageiq/providers/autosde/inventory/persister/storage_manager.rb
```diff
@@ -23,6 +23,29 @@ def initialize_inventory_collections
       )
     end
 
+    # addresses
+    add_collection(physical_infra, :addresses) do |builder|
+      builder.add_default_values(
+        :ems_id => ->(persister) { persister.manager.id }
+      )
+      builder.add_properties(
```
gregoryb-at-ibm, line 31: You can add all the properties in one call to `builder.add_properties`

agrare:
Thoughts on splitting out the openapi client into its own gem?  Feel like we've kicked this can down the road a couple of times

app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
llivne: > You are calling this API twice. Once here and once in physical_storage_consumers. For performance, it's better to only call it once, and then process the results.
> So I suggest you unify the methods `addresses` and `physical_storage_consumers`.

I actually think it's clearer if we have a aeparate method for each table. I agree that the same API is called twice but merging it to a single function looks less clear to me.

llivne:
> Thoughts on splitting out the openapi client into its own gem? Feel like we've kicked this can down the road a couple of times

@borisko123 is working on it

app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
```diff
@@ -26,6 +25,44 @@ def storage_resources
     end
   end
 
+  def san_addresses
+    @san_addresses ||= @manager.autosde_client.StorageHostApi.storage_hosts_get.map do |addresses|
+      addresses_array = []
+
+      addresses.addresses.each do |address|
+        if address.port_type == "ISCSI"
+          port_type = IscsiAddress
```
agrare, line 34: These should be the String `"IscsiAddress"` not the class name

app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
```diff
@@ -26,6 +25,44 @@ def storage_resources
     end
   end
 
+  def san_addresses
+    @san_addresses ||= @manager.autosde_client.StorageHostApi.storage_hosts_get.map do |addresses|
+      addresses_array = []
+
+      addresses.addresses.each do |address|
```
agrare, line 32: Instead of a local array, appending to the array on each loop iteration, and returning the array...this can be simplified by using `#map` 

```
addresses.addresses.map do |address|
  {
    :ems_ref => address.uuid,
    ...
  }
end
```

Map will automatically build an array based on the return from the block and return it

app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
```diff
@@ -26,6 +25,44 @@ def storage_resources
     end
   end
 
+  def san_addresses
+    @san_addresses ||= @manager.autosde_client.StorageHostApi.storage_hosts_get.map do |addresses|
```
agrare, line 29: Since you're making two calls to `@manager.autosde_client.StorageHostApi.storage_hosts_get` maybe we should cache that?

Also a little confusing to call this `addresses` when below you're calling it `host_initiators`

Recommend:
```
def storage_hosts
  @storage_hosts ||= @manager.autosde_client.StorageHostApi.storage_hosts_get
end

def san_addresses
  @san_addresses ||= storage_hosts.map do |storage_host|
  end
end

def host_initiators
  @host_initiators ||= storage_hosts.map ...
  end
end
```

Honestly these look less like "collector" methods and more like "parser" methods.  Can you move `san_addresses` and `host_initiators` to the Parser?  That would leave us with just `def storage_hosts` in the collector

app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
```diff
@@ -43,6 +45,28 @@ def storage_resources
     end
   end
 
+  def san_addresses
+    collector.san_addresses.each do |san_addresses_array|
+      san_addresses_array.each do |san_address_hash|
```
agrare, line 50: Do you need the array of arrays?  Or could you just `.flat_map` in the collector?

app/models/manageiq/providers/autosde/inventory/persister/storage_manager.rb
```diff
@@ -23,6 +23,26 @@ def initialize_inventory_collections
...
+    # host initiators
+    add_collection(physical_infra, :host_initiators) do |builder|
+      builder.add_default_values(
+        :ems_id => ->(persister) { persister.manager.id }
+      )
+      builder.add_properties(
+        :parent_inventory_collections => [:physical_storages]
+      )
+    end
+
```
agrare, line 45: Lets add these to the core inventory collection here: https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/inventory/persister/builder/physical_infra_manager.rb

agrare: Actually shouldn't these all be using the `storage` persister builder not `physical_infra` ??

app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
llivne: fixed

app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
```diff
@@ -26,6 +25,44 @@ def storage_resources
...
+      addresses_array = []
+
+      addresses.addresses.each do |address|
+        if address.port_type == "ISCSI"
```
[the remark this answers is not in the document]
llivne: @agrare I change it to hash, I think its more efficient and easier to work with, what do you think ?

app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
llivne: I changed the code so now you only call the api once

app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
llivne: changed like you suggested, only with `flat_map`

app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
llivne: @agrare, I modified the code to use flat_map.

app/models/manageiq/providers/autosde/inventory/persister/storage_manager.rb
llivne: @agrare they should. I guess this is leftovers we missed since at the beginning we put autosde under `physical_infra`. I changed it to `storage`

llivne: @agrare what is the logic in placing some functionality in the core and some in the persister? why only moving the host_initiator into https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/inventory/persister/builder/storage_manager.rb ? 

if you think we should move more stuff to the core maybe I should do it in a new PR ? or is it ok to keep it inside our provider? (I think that this is functionality that currently only our provider has)

app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
llivne: done

app/models/manageiq/providers/autosde/inventory/persister/storage_manager.rb
agrare: @llivne the idea is that host_initiators are going to be a core model and should be able to be refreshed by other providers.  By having the inventory collection definition in core it allows other providers to reference them without having to redefine the same code you have here.

The issue with physical_infra vs storage is a different one and can be a separate PR

app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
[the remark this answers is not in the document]
agrare: :+1: looks much nicer

app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
```diff
@@ -43,6 +45,54 @@ def storage_resources
...
   end
 
+  def san_addresses
+    port_type = {
+      "ISCSI"  => "IscsiAddress",
+      "FC"     => "FiberChannelAddress",
+      "NVMeFC" => "NvmeAddress"
+    }
+
+    san_addresses = collector.storage_hosts.flat_map do |host_initiator|
```
agrare, line 55: So now you don't even need to map the intermediate parsed hash, you can build it in one shot:
```ruby
# Collector
def storage_host_addresses
  @storage_host_addresses ||= storage_hosts.flat_map(&:addresses)
end

# Parser
collector.storage_host_addresses.each do |address|
  persister.san_addresses.build(
    :ems_ref     => address.uuid,
    :owner       => persister.host_initiators.lazy_find(host_initiator.uuid),
    :type        => port_type[address.port_type],
    :iqn         => address.iqn,
    :chap_name   => address.chap_name,
    :chap_secret => address.chap_secret,
    :wwpn        => address.wwpn
  )
end
```

This saves you from having the whole delete the `:host_initiators_uuid` and add back the whole `**san_address_hash` plus the `:owner` lazy fine

app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
```diff
@@ -43,6 +45,54 @@ def storage_resources
...
+      host_initiators_uuid = san_address_hash.delete(:host_initiators_uuid)
+      persister.san_addresses.build(
+        **san_address_hash,
+        :owner => persister.host_initiators.lazy_find(host_initiators_uuid)
+      )
+    end
+  end
+
+  def host_initiators
+    host_initiators = collector.storage_hosts.map do |host_initiator|
```
agrare, line 79: Same thing here, this can just become

```ruby
collector.storage_hosts.each do |host_initiator|
  persister.host_initiators.build(
    :name             => host_initiator.name,
    :ems_ref          => host_initiator.uuid,
    :physical_storage => persister.physical_storages.lazy_find(host_initiator.storage_system)
  )
end
```

app/models/manageiq/providers/autosde/inventory/persister/storage_manager.rb
llivne: done. see also https://github.com/ManageIQ/manageiq/pull/20858/files#diff-f623f973e49e35b8e6526b8a41057bb3707c90ed14623c932e38054ec081fb3dR21

llivne: done. see also https://github.com/ManageIQ/manageiq/pull/20858/files#diff-f623f973e49e35b8e6526b8a41057bb3707c90ed14623c932e38054ec081fb3dR21

app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
llivne: great. done!

app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
llivne: @agrare  great suggestion, but I still need the info about the host_initiator (not just the address info) for the owner value, so I did it a bit different then how u suggested.

llivne: https://github.com/ManageIQ/manageiq-providers-autosde/pull/49/files#diff-e094665ad11140a16079e84958bf4cbfddefeffd1ee276bc59a7a62e67ddd089R55-R68

agrare: :+1: nice

app/models/manageiq/providers/autosde/inventory/persister/storage_manager.rb
agrare: :+1:

agrare:
@miq-bot cross-repo-tests manageiq-providers-autosde#49 including manageiq#20858

agrare:
@llivne looks like you need to update the refresher VCR to fix the following spec failure https://travis-ci.com/github/ManageIQ/manageiq-cross_repo-tests/builds/212186832#L2935-L2999

agrare:
Cross repo tests are passing now, will merge once https://github.com/ManageIQ/manageiq/pull/20858 is merged

PR #51, opened 2021-01-17
Implement define host initiator

llivne:
Host Initiators (AKA physical storage hosts) are virtual entities that only exists as a set of definitions within the `PhysicalStorage' and hold the information about consumers/hosts that use (defined) the physical storage.

This PR includes the work for adding action of define new host initiator to miq

Screenshot 2021-01-17 063206
Screenshot 2021-01-17 063326

Host-initiators example data
---------
Below are diagrams that shows how we'd represent rather standard scenario. 

All items in white are already part of miq:
We have a single IRL storage-system Called SVC Storage 1.
It has 2 storage pools: Pool1 and Pool2.
Pool1 has two volumes: Volume1 and Volume2.

All items in blue are changes done in this PR:
The Physical Storage has a single host-initiator defined in it. It has been defined with two addresses: iqn-addr-1, and wwpn-addr-1. For the sake of brevity I won't use real IQN and WWPN addresses as they are too long. 

All items in peach are future drafts pending changes (not part of this PR):
The host-initiator in the storage-system has been mapped to Volume1.
A DB server consumes Volume1.
One of the FC addresses of the DB server is wwpn-addr-2. Remeber that this address was defined on the host-initiator above. This, together with the fact that the host-initiator was mapped to Volume1 allows the DB server to consume it.

In this diagram, the addresses are implemented as an STI.
There's a general SAN (Storage Area Network) Address table that holds all possible info for all address. It's expected to be a sparse table as concrete addresses will only populate their specific fields. 
So FcAddresses will have nil for their iscsi values such as IQN and CHAP. And the reverse holds true for IscsiAddresses.

links
-----
https://github.com/ManageIQ/manageiq/pull/20962
https://github.com/ManageIQ/manageiq-api/pull/985
https://github.com/ManageIQ/manageiq-ui-classic/pull/7587


gregoryb-at-ibm (approved):

app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
```diff
@@ -0,0 +1,21 @@
+class ManageIQ::Providers::Autosde::StorageManager::HostInitiator < ::HostInitiator
+  supports :create
+
+  def self.raw_create_host_initiator(_ext_management_system, _options = {})
+    host_initiator_to_create = _ext_management_system.autosde_client.StorageHostCreate(
+      :name           => _options['name'],
```
agrare, line 6: If you're going to use `_options` then it shouldn't have a leading `_` (that indicates that the parameter isn't going to be used)

app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
```diff
@@ -0,0 +1,21 @@
...
+      :port_type      => _options['port_type'],
+      :storage_system => PhysicalStorage.find(_options['physical_storage_id']).ems_ref,
+      :iqn            => _options['iqn'] || "",
+      :wwpn           => _options['wwpn'] || "",
+      :chap_name      => _options['chap_name'] || "",
+      :chap_secret    => _options['chap_secret'] || ""
```
agrare, line 12: I assume we're going to need DDF forms for these params?  cc @skateman

app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
```diff
@@ -0,0 +1,21 @@
+class ManageIQ::Providers::Autosde::StorageManager::HostInitiator < ::HostInitiator
+  supports :create
+
+  def self.raw_create_host_initiator(_ext_management_system, _options = {})
+    host_initiator_to_create = _ext_management_system.autosde_client.StorageHostCreate(
```
agrare, line 5: Should be just `ext_management_system`

app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
```diff
@@ -0,0 +1,21 @@
...
+      :iqn            => _options['iqn'] || "",
+      :wwpn           => _options['wwpn'] || "",
+      :chap_name      => _options['chap_name'] || "",
+      :chap_secret    => _options['chap_secret'] || ""
+    )
+
+    begin
+      _ext_management_system.autosde_client.StorageHostApi.storage_hosts_post(host_initiator_to_create)
+    ensure
+      EmsRefresh.queue_refresh(_ext_management_system)
```
agrare, line 18: You sure you want to refresh even if this fails?  I would have thought the opposite, only refresh if the create succeeded.

Is there a timing issue here?  Is the storage_host guaranteed to be created once this API call returns or does it just initiate a process to create it?

agrare (changes requested):

app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
```diff
@@ -0,0 +1,21 @@
+class ManageIQ::Providers::Autosde::StorageManager::HostInitiator < ::HostInitiator
+  supports :create
+
+  def self.raw_create_host_initiator(_ext_management_system, _options = {})
+    host_initiator_to_create = _ext_management_system.autosde_client.StorageHostCreate(
+      :name           => _options['name'],
+      :port_type      => _options['port_type'],
+      :storage_system => PhysicalStorage.find(_options['physical_storage_id']).ems_ref,
+      :iqn            => _options['iqn'] || "",
```
agrare, line 9: Its a little strange that these have to be empty strings not `nil` if they aren't present in the options.

Are these values actually required by the API?

app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
skateman: I don't really get the context, but probably...

llivne: @agrare @skateman I did the DDF forms, you can look on it in https://github.com/ManageIQ/manageiq-ui-classic/pull/7587/files#diff-9b1b0595c256080f1cd7cbc9d1d01f2d32608b561cf00f0667b4a76afc8f8e35R1-R115

app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
llivne: @agrare these values are required (this was decided by the designer). I actually tried to pass nil at first but it didnt work (why ????), so I had to use empty string.

app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
llivne: fixed

app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
llivne: fixed

app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
agrare: Yeah that's what I figured, I don't know that `""` is a good workaround if the API expects this to be present.  We might want ot make all of these properties required if the API requires it

app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
agrare: @llivne I still don't think you want to refresh the provider if creating the storage_host fails since there would be no change in the inventory.

llivne: Agree. Fixed.

app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
llivne: chap_name and secret are optional fields and only relevant for ISCSI and not for FC so I dont think we should set them as required. I know that autosde does required them but I think this is something we should reconsider. @gregoryb-at-ibm what do you think?

agrare (approved):
We can work out the required options in a follow-up.

app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
agrare: > chap_name and secret are optional fields and only relevant for ISCSI and not for FC

But does the API fail for FC if chap_name or secret are `nil` ?

llivne: yes, it must get an empty string

agrare: That's weird then, if it isn't required for the type it shouldn't fail if it isn't present IMO but if this is what we have to do to work around the API that's okay

llivne: I agree with you, and I'm going to challenge our current autosde design again (i tried a few weeks ago but maybe I wasnt clear enough). For now the work around is required, but if we change autosde api in the future I will do the required modifications in miq.

agrare: :+1:

Issue #10, opened 2020-10-02
Extract OpenapiClient to an sdk client gem


agrare:
It would be nice to extract the OpenapiClient to its own autosde_oas SDK gem

gregoryb-at-ibm:
@agrare , Is this a blocker? Could we move this to a gem at a later stage? I'm concerned that we are making big changes, but we can't really do functional tests until we finish this initial round of PRs.

Also, I'm not familiar with what is an "SDK gem"? Do you have an example of such a thing? How is it different from a regular gem?

agrare:
Not a blocker

agrare:
@gregoryb-at-ibm what you have in `OpenapiClient` is an sdk gem, it has a [gemspec](https://github.com/Autosde/manageiq-providers-autosde/blob/master/app/models/manageiq/providers/autosde/storage_manager/openapi_client/generated/openapi_client.gemspec) and everything

https://github.com/aws/aws-sdk-ruby and https://github.com/vmware/rbvmomi/ are other examples of SDK gems, just a way of saying the library that talks to the provider's API

borisko123:
@agrare this is closed by https://github.com/ManageIQ/manageiq-providers-autosde/pull/52
Can we close this issue?

agrare:
:+1: yes

PR #52, opened 2021-01-19
replace autosde openAPI ( OAS) files with gem in Rubygem

borisko123:
This PR replace autosde OpenapiClient with its own autosde_oas SDK gem
for more info look inside: https://github.com/ManageIQ/manageiq-providers-autosde/issues/10

links
-----
https://github.com/ManageIQ/manageiq/pull/20966


manageiq-providers-autosde.gemspec
```diff
@@ -19,6 +19,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "typhoeus", "~> 1.4"
+  spec.add_dependency "autosde_openapi_client", "~> 1.0"
```
agrare, line 22: :+1: nice there you go

Not sure if you're sold on the name but having `openapi` in the gem name probably isn't needed.  Maybe `autosde_client` or `autosde_oas_client` or something like that?

llivne: @borisko123 I added autosde_openapi_client with fixed version to autosde genspec (so the changes in the gemfile in the core are not needed). please make sure everything is working, and also consider @agrare suggestions on changing the name and using "~> 1.0" for the version (I used the fixed version because I remember you are still working on 1.0.4 and I'm not sure if it will work with it).

llivne:
@borisko123 some specs fails (see travis). Please update them to use the new gem

manageiq-providers-autosde.gemspec
```diff
@@ -18,6 +18,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
+  spec.add_dependency "autosde_openapi_client", "=1.0.3"
```
Fryguy, line 21: Prefer

```suggestion
  spec.add_dependency "autosde_openapi_client", "~> 1.0"
```

This will allow patch-level and minor-level updates to automatically be pulled in, so you don't have to constantly be keeping this number up to date.

manageiq-providers-autosde.gemspec
borisko123: @agrare please let me remain the name 'autosde_openapi_client' for gem.
1. it is not so simple to change it 
2. the oas means exactly OpenAPI Specification (OAS), which is, in essence, the same
3. oas also means https://en.wikipedia.org/wiki/Organization_of_American_States, may be confused😄

borisko123:
> @borisko123 some specs fails (see travis). Please update them to use the new gem

done

llivne:
@borisko123 I approve, depending on @agrare response regarding the gem name

llivne (approved):

app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
```diff
@@ -22,10 +22,10 @@ def initialize(username:, password:, host:, port: 443, token: nil, scheme: 'http
     super(configure_openapi_client)
   end
 
-  # generates access methods for OpenApiClient , like StorageSystemApi or CreateVolume
+  # generates access methods for AutosdeOpenapiClient , like StorageSystemApi or CreateVolume
   # usage : <this>.StorageSystemApi.storage_systems_get
-  OpenapiClient.constants.each do |method|
-    clazz = "OpenapiClient::#{method}".constantize
+  AutosdeOpenapiClient.constants.each do |method|
+    clazz = "AutosdeOpenapiClient::#{method}".constantize
```
gregoryb-at-ibm, line 28: We eventually want to move all these patches to the gem. And also the specs.

spec/models/manageiq/providers/autosde/autosde_client_all_storage_models_spec.rb
```diff
@@ -10,7 +10,7 @@
   # Using client.VolumeCreat(service: service, name: vol_name, size: 10)
```
gregoryb-at-ibm, line 10: We will eventually want to move the client specs to the client gem

gregoryb-at-ibm (approved):

manageiq-providers-autosde.gemspec
agrare: Okay I see it is already on rubygems with this name, I don't like it because the fact that it is an openapi client doesn't matter to the caller it is an implementation detail.

app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
```diff
@@ -1,8 +1,8 @@
-require 'autosde_oas_client/generated/lib/autosde_openapi_client'
+require 'autosde_openapi_client'
 
-class ManageIQ::Providers::Autosde::StorageManager::AutosdeClient < OpenapiClient::ApiClient
+class ManageIQ::Providers::Autosde::StorageManager::AutosdeClient < AutosdeOpenapiClient::ApiClient
   include Vmdb::Logging
-  include OpenapiClient
+  include AutosdeOpenapiClient
```
agrare, line 5: Hm what is this `include` actually doing for us?  This module isn't really doing anything besides defining a `.configure` method on `AutosdeOpenapiClient` which it doesn't look like you're using

app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
```diff
@@ -89,6 +89,9 @@ def configure_openapi_client
       config.host = @host
```
agrare, line 89: [this](https://github.com/ManageIQ/manageiq-providers-autosde/pull/52/files#diff-c4817412fc012e61a1492a4a42c07e4cbdc72dcc53a76c759a80a848ceb68415R86) should be changed to `AutosdeOpenapiClient::Configuration`

app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
agrare: Agreed, the override for call_api and auto-login looks a little hacky also.  Best to login at the start of a refresh or other operation and move on.  But we can tackle this in a follow-up.

app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
agrare: Believe this is working because of the include but that's not really the right way.  I'm going to merge this so we can move forward and I'll put in a follow-up PR to fix this

manageiq-providers-autosde.gemspec
gregoryb-at-ibm: I agree. Maybe we'll create a differently named gem at a later phase.

app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
agrare: https://github.com/ManageIQ/manageiq-providers-autosde/pull/53

PR #57, opened 2021-02-09
Creating volume mapping

gregoryb-at-ibm:
Needs to be done after  https://github.com/ManageIQ/manageiq-providers-autosde/pull/56

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/volume_mapping.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/volume_mapping.rb
@@ -1,0 +1,14 @@
+class ManageIQ::Providers::Autosde::StorageManager::VolumeMapping < ::VolumeMapping
+  supports :create
+
+  def self.raw_create_volume_mapping(ext_management_system, options = {})
+
+    volume_mapping_to_create = ext_management_system.autosde_client.StorageHostVolumeMappingCreate(
+      :host => HostInitiator.find(options['host_initiator_id']).ems_ref,
+      :volume => CloudVolume.find(options['cloud_volume_id']).ems_ref
+    )
+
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/autosde/storage_manager/volume_mapping.rb
```diff
@@ -0,0 +1,14 @@
+class ManageIQ::Providers::Autosde::StorageManager::VolumeMapping < ::VolumeMapping
```
agrare, line 1: @gregoryb-at-ibm I just noticed that the volume_mappings table doesn't have a `:type` column so STI isn't possible.

Can you put in a PR to add that column to volume_mappings?

agrare (changes requested):

app/models/manageiq/providers/autosde/storage_manager/volume_mapping.rb
gregoryb-at-ibm: Ah! Must have missed it...
How come it still works and I can create volume-mapping?
And how come we don't have `:type` in `host_initiators`?

Here's the schema PR https://github.com/ManageIQ/manageiq-schema/pull/558

agrare: > How come it still works and I can create volume-mapping?

Probably because you're using class_by_ems which is returning a valid class and calling a method off of it, but that class isn't actually backed by the database record.

E.g.:
```
host_initiator = HostInitiator.class_by_ems(ems)
host_initiator.raw_create_volume # "works"
HostInitiator.find(:id).raw_create_volume # would blow up
```

If you have a subclass for `HostInitiator` then it also needs a type

agrare (approved):

PR #59, opened 2021-02-17
Removed default for security protocol when creating new EMS

gregoryb-at-ibm:
The user now needs to explicitly choose their security protocol when creating a new Autosde EMS.

PR #60, opened 2021-02-17
Adjust empty/initial values for select fields in DDF schemas

skateman:
Because carbon has four different dropdowns, the DDF schemas need to have explicit initialValue or includeEmpty settings in order to make the form validation work.

@miq-bot assign @agrare

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager.rb
@@ -224,13 +224,14 @@ class ManageIQ::Providers::Autosde::StorageManager < ManageIQ::Providers::Storag
               :validationDependencies => %w[type endpoints.default.hostname authentications.default.userid authentications.default.password],
               :fields                 => [
                 {
-                  :component  => "select",
-                  :id         => "endpoints.default.security_protocol",
-                  :name       => "endpoints.default.security_protocol",
-                  :label      => _("Security Protocol"),
-                  :isRequired => true,
-                  :validate   => [{:type => "required"}],
-                  :options    => [
+                  :component    => "select",
+                  :id           => "endpoints.default.security_protocol",
+                  :name         => "endpoints.default.security_protocol",
+                  :label        => _("Security Protocol"),
+                  :isRequired   => true,
+                  :validate     => [{:type => "required"}],
+                  :initialValue => 'ssl-no-validation',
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/autosde/storage_manager.rb
```diff
@@ -224,13 +224,14 @@ def self.params_for_create
...
-                  :isRequired => true,
-                  :validate   => [{:type => "required"}],
-                  :options    => [
+                  :component    => "select",
+                  :id           => "endpoints.default.security_protocol",
+                  :name         => "endpoints.default.security_protocol",
+                  :label        => _("Security Protocol"),
+                  :isRequired   => true,
+                  :validate     => [{:type => "required"}],
+                  :initialValue => 'ssl-no-validation',
```
agrare, line 233: @gregoryb-at-ibm it seems like you don't want an initial value here based on https://github.com/ManageIQ/manageiq-providers-autosde/pull/59 is that right?

gregoryb-at-ibm: I actually created #59 because the initial value stopped working. I tried fixing it by adding `initialValue='ssl-no-validation'` like in this PR. But that didn't fix the issue. The form still wouldn't validate for me.

I decided that since we are going to add 2 more options soon anyway, and when we do we won't want a default, it's OK to just remove the default now.

We can try working together to try and figure out why it's not working for me if you want.

skateman:
Closing in favor of https://github.com/ManageIQ/manageiq-providers-autosde/pull/59

app/models/manageiq/providers/autosde/storage_manager.rb
gregoryb-at-ibm: Here's how it looks like when I use `initialValue`

image

image

gregoryb-at-ibm: @Fryguy 
FYI ^^
This might mean there's a bug in how `initialValue` works.

skateman: This is actually a bug in Carbon itself, it implicitly selects a field but doesn't tell the validator. According to the DDF people: "no one can implement a dropdown properly" :laughing:

PR #66, opened 2021-04-12
Update autosde provider to use rails secrets

agrare:
Add pluggable provider secrets and convert the Rspec.configuration over to the more standard use (at least by the rest of MIQ) of Rails.application.secrets

https://github.com/ManageIQ/manageiq/issues/21157


agrare:
Thanks @gregoryb-at-ibm !

PR #72, opened 2021-07-15
Add support for editing autosde cloud volume

llivne:
This PR is part of PRs that solve Autosde cloud volume editing

image

If specific info about a cloud volume is passed then we are under update mode (not creation) and the update field should be used.

Links
----------------
https://github.com/ManageIQ/manageiq-ui-classic/pull/7788
https://github.com/ManageIQ/manageiq-api/pull/1056

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
@@ -57,10 +57,6 @@ class ManageIQ::Providers::Autosde::StorageManager::CloudVolume < ::CloudVolume
   end
 
   def params_for_update
-    ems = ExtManagementSystem.find(ems_id)
-    storage_service_id = ems.cloud_volumes.find_by(:id => id)[:storage_service_id]
-    service_value = ems.storage_services.find_by(:id => storage_service_id)[:name]
-
     {
       :fields => [
         {
@@ -70,7 +66,7 @@ class ManageIQ::Providers::Autosde::StorageManager::CloudVolume < ::CloudVolume
           :label        => _("Storage Pool"),
           :isRequired   => true,
           :validate     => [{:type => "required"}],
-          :initialValue => service_value,
+          :initialValue => storage_service.name,
           :isDisabled   => true
         },
         {
--- a/app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
@@ -8,7 +8,7 @@ class ManageIQ::Providers::Autosde::StorageManager::CloudVolume < ::CloudVolume
   def self.raw_create_volume(ext_management_system, options = {})
     # @type [StorageService]
     vol_to_create = ext_management_system.autosde_client.VolumeCreate(
-      :service => ExtManagementSystem.find(options["ems_id"]).storage_services.detect {|e| e.id.to_s ==  options["storage_service_id"]}.ems_ref,
+      :service => ExtManagementSystem.find(options["ems_id"]).storage_services.detect { |e| e.id.to_s == options["storage_service_id"] }.ems_ref,
       :name    => options["name"],
       :size    => options["size"]
     )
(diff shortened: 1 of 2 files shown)


llivne:
I updated the code to work with OPTIONS. This is now possible because of @agrare modifications: ManageIQ/manageiq-api#1056 (comment)

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
```diff
@@ -8,7 +8,7 @@ class ManageIQ::Providers::Autosde::StorageManager::CloudVolume < ::CloudVolume
   def self.raw_create_volume(ext_management_system, options = {})
     # @type [StorageService]
     vol_to_create = ext_management_system.autosde_client.VolumeCreate(
-      :service => options["storage_service"],
+      :service => ExtManagementSystem.find(options["ems_id"]).storage_services.detect { |e| e.id.to_s == options["storage_service_id"] }.ems_ref,
```
agrare, line 11: If you're passed in the `ext_management_system` then you don't need to `ExtManagementSystem.find(options["ems_id"])`

agrare: > `.storage_services.detect { |e| e.id.to_s == options["storage_service_id"] }.ems_ref`

This looks like it can just be `ext_management_system.storage_services.find_by(:id => options["storage_service_id"])&.ems_ref`

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
```diff
@@ -58,19 +56,52 @@ def raw_safe_delete_volume
     EmsRefresh.queue_refresh(ext_management_system)
   end
 
-  def self.params_for_create(provider)
-    services = provider.storage_services.map { |service| {:value => service.ems_ref, :label => service.name} }
+  def params_for_update
+    ems = ExtManagementSystem.find(ems_id)
+    service_value = ems.storage_services.detect { {:value => id, :label => name} }[:name]
```
agrare, line 61: I don't think this is doing anything `.detect { {:value => id, :label => name} }`  if anything it is likely just returning the first service always because `{:value => id, :label => name}` is truthy

Not sure what you're trying to do, but is it `ems.storage_services.find_by(:value => id, :label => name)` ?

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
llivne: done

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
llivne: It was suppose to be:
`ems = ExtManagementSystem.find(ems_id)`
`storage_service_id = ems.cloud_volumes.find_by(:id => id)[:storage_service_id]`
`service_value = ems.storage_services.find_by(:id => storage_service_id)[:name]`

I pushed the correct code now

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
```diff
@@ -58,19 +56,53 @@ def raw_safe_delete_volume
     EmsRefresh.queue_refresh(ext_management_system)
   end
 
-  def self.params_for_create(provider)
-    services = provider.storage_services.map { |service| {:value => service.ems_ref, :label => service.name} }
+  def params_for_update
+    ems = ExtManagementSystem.find(ems_id)
+    storage_service_id = ems.cloud_volumes.find_by(:id => id)[:storage_service_id]
```
agrare, line 61: The preferred way to reference an attribute on an object is with `.storage_service_id` rather than `[:storage_service_id]`

agrare: Actually, `ems.cloud_volumes.find_by(:id => id)` is the same thing as `self` since you're in an instance method, so you just need to call `storage_service_id`.

Since `CloudVolume` has `belongs_to :storage_service` you can actually drop the explicity `find_by(:id => storage_service_id` and let rails do that query.

This whole section can become
```ruby
service_value = storage_service.name
```

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
```diff
@@ -58,19 +56,53 @@ def raw_safe_delete_volume
...
+    service_value = ems.storage_services.find_by(:id => storage_service_id)[:name]
+
     {
       :fields => [
         {
-          :component    => "select",
+          :component    => "text-field",
```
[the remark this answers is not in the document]
llivne: we cant change the pool, so this field is disabled, and only shown for information. I can use select but then I need to populate it. the code will look like that:
```
    ems = ExtManagementSystem.find(ems_id)
    services = ems.storage_services.map { |service| {:value => service.id, :label => service.name} }

    {
      :fields => [
        {
          :component    => "select",
          :name         => "storage_service",
          :id           => "storage_service",
          :label        => _("Storage Pool"),
          :isRequired   => true,
          :validate     => [{:type => "required"}],
          :options      => services,
          :initialValue => storage_service.id,
          :isDisabled   => true
        },
```

the final outcome will be the same if we use select or text-field, the name of the pool disabled.
I think that text-field is better for this scenario, but If you prefer select let me know and I will change it.

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
llivne: good suggestion. fixed

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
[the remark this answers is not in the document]
agrare: > we cant change the pool, so this field is disabled,

Oh, okay interesting this is fine then

agrare (approved):

ereztheodorou:
@miq-bot add_label morphy/yes?

Fryguy:
Backported to `morphy` in commit 517a34e5962e022b1cc808738440baafd989e318.

```text
commit 517a34e5962e022b1cc808738440baafd989e318
Author: Adam Grare <adam@grare.com>
Date:   Thu Aug 5 07:51:40 2021 -0400

    Merge pull request #72 from Autosde/fix_edit_cloud_volume
    
    Add support for editing autosde cloud volume
    
    (cherry picked from commit 1b37b0d3eb2af68b6bad130d32535bd9462875c3)
```

PR #75, opened 2021-08-08
added health state to collector

IlyaKessel:
Added health_state to the collector in physical storage and cloud volume objects
this PR depend on https://github.com/ManageIQ/manageiq-schema/pull/598

image

image


agrare:
@IlyaKessel can you look at the travis failures?  Looks like the VCR might have to be re-recorded

agrare (approved):

Issue #61, opened 2021-02-18
Cleanup redundant validate method from #58


gregoryb-at-ibm:
Cleanup redundant validate method in https://github.com/ManageIQ/manageiq-providers-autosde/pull/58

chessbyte:
@gregoryb-at-ibm has this been addressed or still a to-do item?

gregoryb-at-ibm:
@chessbyte , this hasn't been fixed. 
I'll bump it in our internal issue-tracker.

gregoryb-at-ibm:
Closed by @chessbyte  at https://github.com/ManageIQ/manageiq-providers-autosde/pull/78

PR #78, opened 2021-08-23
Convert cloud volume operations to SupportsFeatureMixin

chessbyte:
Part of https://github.com/ManageIQ/manageiq/pull/21393

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
@@ -1,8 +1,13 @@
 class ManageIQ::Providers::Autosde::StorageManager::CloudVolume < ::CloudVolume
   supports :create
-  supports :safe_delete
+  supports :safe_delete do
+    unsupported_reason_add(:safe_delete, _("the volume is not connected to an active provider")) unless ext_management_system
+  end
   supports :update do
-    unsupported_reason_add(:update, _("The Volume is not connected to an active Provider")) unless ext_management_system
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
```diff
@@ -1,8 +1,13 @@
...
   supports :create
-  supports :safe_delete
+  supports :safe_delete do
+    unsupported_reason_add(:safe_delete, _("the volume is not connected to an active provider")) unless ext_management_system
+  end
   supports :update do
-    unsupported_reason_add(:update, _("The Volume is not connected to an active Provider")) unless ext_management_system
+    unsupported_reason_add(:update, _("the volume is not connected to an active provider")) unless ext_management_system
+  end
+  supports :delete_volume do
```
agrare, line 9: Think this should be `supports :delete` since that is the common pattern for crud operations (`:create`, `:update`, `:delete`)

gregoryb-at-ibm:
Hi @chessbyte ,
Looks like this would also resolve https://github.com/ManageIQ/manageiq-providers-autosde/issues/61 ?

gregoryb-at-ibm (approved):

agrare (approved):

chessbyte:
@gregoryb-at-ibm next time, please be mindful that this PR is part of a set described in the [OP](https://github.com/ManageIQ/manageiq-providers-autosde/pull/78#issue-717968382) and needs to be merged together with its dependencies.  As such, this may cause temporary issues until the rest of the set is reviewed and merged.  /cc @agrare

PR #107, opened 2021-11-21
add targeted refresh for editing cloud-volumes and physical-storages

llivne:
replacing full refresh with targeted refresh for editing cloud volumes and physical storages.

### DIFF
--- a/app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
+++ b/app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
@@ -4,30 +4,11 @@ class ManageIQ::Providers::Autosde::Inventory::Collector::StorageManager < Manag
   attr_accessor :manager
 
   def physical_storages
-    @physical_storages ||= @manager.autosde_client.StorageSystemApi.storage_systems_get.map
-
-    @physical_storages.map do |system|
-      {
-        :name             => system.name,
-        :ems_ref          => system.uuid,
-        :system_type_uuid => system.system_type.uuid,
-        :health_state     => system.status
-      }
-    end
+    @physical_storages ||= @manager.autosde_client.StorageSystemApi.storage_systems_get
   end
 
   def storage_resources
-    @storage_resources ||= @manager.autosde_client.StorageResourceApi.storage_resources_get.map
-
-    @storage_resources.map do |resource|
-      {
-        :name                => resource.name,
-        :ems_ref             => resource.uuid,
-        :logical_free        => resource.logical_free,
-        :logical_total       => resource.logical_total,
-        :storage_system_uuid => resource.storage_system
-      }
-    end
+    @storage_resources ||= @manager.autosde_client.StorageResourceApi.storage_resources_get
   end
 
   def storage_hosts
@@ -35,71 +16,23 @@ class ManageIQ::Providers::Autosde::Inventory::Collector::StorageManager < Manag
   end
 
   def host_volume_mappings
-    @host_volume_mappings ||= @manager.autosde_client.StorageHostVolumeMappingApi.storage_hosts_mapping_get.map
-
-    @host_volume_mappings.map do |mapping|
-      {
-        :lun                 => mapping.lun,
-        :host_initiator_uuid => mapping.host,
-        :volume_uuid         => mapping.volume,
-        :ems_ref             => mapping.uuid,
-      }
-    end
+    @host_volume_mappings ||= @manager.autosde_client.StorageHostVolumeMappingApi.storage_hosts_mapping_get
   end
 
   def cluster_volume_mappings
-    @cluster_volume_mappings ||= @manager.autosde_client.HostClusterVolumeMappingApi.host_clusters_mapping_get.map
-
-    @cluster_volume_mappings.map do |mapping|
-      {
-        :lun                       => mapping.lun,
-        :host_initiator_group_uuid => mapping.cluster,
-        :volume_uuid               => mapping.volume,
-        :ems_ref                   => mapping.uuid,
-      }
-    end
+    @cluster_volume_mappings ||= @manager.autosde_client.HostClusterVolumeMappingApi.host_clusters_mapping_get
   end
 
   def cloud_volumes
-    # Get all of the cloud volumes if specific cloud volumes were not set (e.g. with targeted refresh)
-    @cloud_volumes ||= @manager.autosde_client.VolumeApi.volumes_get.map
-
-    @cloud_volumes.map do |volume|
-      {
-        :name                  => volume.name,
-        :size                  => volume.size * 1024 * 1024 * 1024,
-        :ems_ref               => volume.uuid,
-        :storage_resource_uuid => volume.storage_resource,
-        :storage_service_uuid  => volume.service,
-        :status                => volume.component_state,
-        :health_state          => volume.status
-      }
-    end
+    @cloud_volumes ||= @manager.autosde_client.VolumeApi.volumes_get
   end
 
   def storage_services
-    @storage_services ||= @manager.autosde_client.ServiceApi.services_get.map
-
-    @storage_services.map do |service|
-      {
-        :name        => service.name,
-        :description => service.description,
-        :version     => service.version,
-        :ems_ref     => service.uuid
-      }
-    end
+    @storage_services ||= @manager.autosde_client.ServiceApi.services_get
   end
 
   def physical_storage_families
-    @physical_storage_families ||= @manager.autosde_client.SystemTypeApi.system_types_get.map
-
-    @physical_storage_families.map do |system_type|
-      {
-        :name    => system_type.name,
-        :ems_ref => system_type.uuid,
-        :version => system_type.version
-      }
-    end
+    @physical_storage_families ||= @manager.autosde_client.SystemTypeApi.system_types_get
   end
 
   def wwpn_candidates
--- a/app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
+++ b/app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
@@ -4,7 +4,9 @@ class ManageIQ::Providers::Autosde::Inventory::Collector::StorageManager < Manag
   attr_accessor :manager
 
   def physical_storages
-    @physical_storages ||= @manager.autosde_client.StorageSystemApi.storage_systems_get.map do |system|
+    @physical_storages ||= @manager.autosde_client.StorageSystemApi.storage_systems_get.map
+
+    @physical_storages.map do |system|
       {
         :name             => system.name,
         :ems_ref          => system.uuid,
@@ -15,7 +17,9 @@ class ManageIQ::Providers::Autosde::Inventory::Collector::StorageManager < Manag
   end
 
   def storage_resources
-    @storage_resources ||= @manager.autosde_client.StorageResourceApi.storage_resources_get.map do |resource|
+    @storage_resources ||= @manager.autosde_client.StorageResourceApi.storage_resources_get.map
+
+    @storage_resources.map do |resource|
       {
         :name                => resource.name,
         :ems_ref             => resource.uuid,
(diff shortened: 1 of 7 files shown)


agrare:
Hey @llivne I think all of the hoops that you're jumping through in the collector is indicative of an issue I brought up a while ago.

You're doing the parsing in the Collector not the Parser.  This would be a good time to fix that before getting too much further into this "set the ivar, call super, ||= the ivar, parse"

app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
```diff
@@ -4,7 +4,9 @@ class ManageIQ::Providers::Autosde::Inventory::Collector::StorageManager < Manag
   attr_accessor :manager
 
   def physical_storages
-    @physical_storages ||= @manager.autosde_client.StorageSystemApi.storage_systems_get.map do |system|
+    @physical_storages ||= @manager.autosde_client.StorageSystemApi.storage_systems_get.map
```
agrare, line 7: `.map` without a block is a little strange, if you just want to turn an enumerator into an array you can use `.to_a`

llivne: I moved all the parser code from the collector into the parser. this is not an issue anymore

llivne:
> Hey @llivne I think all of the hoops that you're jumping through in the collector is indicative of an issue I brought up a while ago.
> 
> You're doing the parsing in the Collector not the Parser. The main purpose for how the collector/parser are split up the way that they are is so that the parser is completely reusable for fulls and targeted refreshes.
> 
> This would be a good time to fix that before getting too much further into this "set the ivar, call super, ||= the ivar, parse"

@agrare I moved all the parser code from the collector into the parser.

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
```diff
@@ -38,13 +38,13 @@ def raw_update_volume(options = {})
       :size => options["size_GB"]
     )
     ext_management_system.autosde_client.VolumeApi.volumes_pk_put(ems_ref, update_details)
-    EmsRefresh.queue_refresh(ems)
+    cloud_vol = ext_management_system.cloud_volumes.find_by(:ems_ref=>ems_ref)
+    EmsRefresh.refresh(cloud_vol)
```
agrare, line 42: You actually don't need to do another query to find the cloud-volume, since this is an instance method `self` already is the record that you're looking for.

Also you don't want to do `EmsRefresh.refresh` because that will run it in the current worker (probably a generic worker) rather than the refresh worker.  All refreshes have to be run by the refresh worker in order to ensure data consistency.

```suggestion
    EmsRefresh.queue_refresh(self)
```

app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
```diff
@@ -21,7 +21,8 @@ def raw_update_physical_storage(options = {})
       :management_ip => options['management_ip'] || ""
     )
     ext_management_system.autosde_client.StorageSystemApi.storage_systems_pk_put(ems_ref, update_details)
-    EmsRefresh.queue_refresh(ext_management_system)
+    ph_storage = ext_management_system.physical_storages.find_by(:ems_ref=>ems_ref)
+    EmsRefresh.refresh(ph_storage)
```
agrare, line 25: Same comment here
```suggestion
    EmsRefresh.queue_refresh(self)
```

app/models/manageiq/providers/autosde/inventory/collector/target_collection.rb
```diff
@@ -8,49 +8,62 @@ def initialize(_manager, _target)
   def physical_storages
     return [] if references(:physical_storages).blank?
 
-    super
+    @physical_storages ||= references(:physical_storages).map do |ems_ref|
+      @manager.autosde_client.StorageSystemApi.storage_systems_get.map { |s| s if s.uuid == ems_ref }.compact.first
```
agrare, line 12: This looks like it should be a `.select` rather than `map`.  Select does exactly what you're trying to do here with the `s if s.uuid == ems_ref` block.

```suggestion
      @manager.autosde_client.StorageSystemApi.storage_systems_get.select { |s| s.uuid == ems_ref }
```

agrare, line 12: Also this looks like it will do an API call per target, is there a way to issue a `GET` on a single resource id (or collection of resource ids) without returning the whole collection?

If you do have to do the whole collection then best to only do that once and then filter down the refs like:
```suggestion
    @physical_storages ||= @manager.autosde_client.StorageSystemApi.storage_systems_get.select { |s| references(:physical_storages).include?(s.uuid) }
```

app/models/manageiq/providers/autosde/inventory/persister/storage_manager.rb
```diff
@@ -1,29 +1,43 @@
...
 
   def initialize_inventory_collections
-    # physical storages
-    add_collection(storage, :physical_storages) do |builder|
-      builder.add_default_values(:ems_id => ->(persister) { persister.manager.id })
-      builder.add_properties(
-        :model_class => ManageIQ::Providers::Autosde::StorageManager::PhysicalStorage
-      )
-    end
+    collect_san_addresses
```
agrare, line 9: `collect_` is a bit confusing because this is in the persister not the collector.  If you have provider overrides the typical pattern is to call this `add_host_initiators`

app/models/manageiq/providers/autosde/inventory/persister/storage_manager.rb
```diff
@@ -1,29 +1,43 @@
...
+      :targeted => targeted?,
+      :parent   => parent
+    }
+  end
+
+  protected
+
+  # san addresses
+  def collect_san_addresses
     add_collection(storage, :san_addresses) do |builder|
```
agrare, line 41: Not for this PR but most of these could move to the core inventory collection https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/inventory/persister/builder/storage_manager.rb

If all you are doing is setting the `ems_id`, `model_class`, and `parent_inventory_collections` that actually isn't provider specific but would be the same for all storage_managers.

app/models/manageiq/providers/autosde/inventory/persister/target_collection.rb
```diff
@@ -2,4 +2,11 @@ class ManageIQ::Providers::Autosde::Inventory::Persister::TargetCollection < Man
   def targeted?
     true
   end
+
+  # Overwrite the base initialize_inventory_collections function, and add to collections only the targeted objects.
+  def initialize_inventory_collections
+    target.manager_refs_by_association.each_key do |target|
+      send("collect_#{target}")
+    end
```
agrare, line 10: Oh now I see why you split all those up into methods.  So the persister/inventory_refresh code for targeted refreshes automatically skips "empty" inventory collections based on the targeted_scope.

This means you don't have to add only collections that are present, if you have a cloud_volume target and no physical_storages the physical_storages collection will be skipped rather than deleting them all.  That is the primary difference between full and targeted refreshes.

app/models/manageiq/providers/autosde/inventory/collector/target_collection.rb
```diff
@@ -8,49 +8,51 @@ def initialize(_manager, _target)
...
 
-    super
+    @storage_resources ||= @manager.autosde_client.StorageResourceApi.storage_resources_get.select { |s| references(:storage_resources).include?(s.uuid) }
   end
 
   def storage_hosts
     return [] if references(:storage_hosts).blank?
-
-    super
   end
```
agrare, line 22: For these that aren't implemented yet best to just return `[]` since currently this will return `nil` if there are any storage_host targets

app/models/manageiq/providers/autosde/inventory/collector/target_collection.rb
```diff
@@ -8,49 +8,51 @@ def initialize(_manager, _target)
...
 
-  def volume_mappings
+  def host_volume_mappings
     return [] if references(:volume_mappings).blank?
+  end
```
agrare, line 26: Same here

app/models/manageiq/providers/autosde/inventory/collector/target_collection.rb
```diff
@@ -8,49 +8,51 @@ def initialize(_manager, _target)
...
 
-    super
+  def cluster_volume_mappings
+    return [] if references(:volume_mappings).blank?
   end
```
agrare, line 30: And here

agrare:
Looks good @llivne just that one issue with the targeted collector returning `nil` then I think everything else can be done in a follow-up

We should add some specs as well, targeted refresh specs are typically done with mocks rather than full VCRs (e.g. do full refresh with the VCR then stub one of the API calls for a targeted refresh).

agrare:
I have some follow-up refactorings that I don't want to conflict with your PR here so I'm going to merge this then follow-up with some more changes

llivne:
> I have some follow-up refactorings that I don't want to conflict with your PR here so I'm going to merge this then follow-up with some more changes

@agrare,thanks for all of your help and feedback. I'm go over all of the latest changes, and make sure all is working ok.

llivne:
@agrare two issues:
1) The queue_refresh didnt worked. It raised "Unknown target type" warning. Adding PhysicalStorage and CloudVolume inside ems_refresh.rb resolved it:
```
if ManageIQ::Providers::Inventory.persister_class_for(target_class).blank? &&
         [VmOrTemplate, Host, PhysicalServer, PhysicalStorage, CloudVolume, ExtManagementSystem, InventoryRefresh::Target].none? { |k| target_class <= k }
        _log.warn("Unknown target type: [#{target_class}].")
        next
      end
```

2) It seems that targeted refreshes DOES NOT automatically skips "empty" inventory collections based on the targeted_scope.
I get errors about not related inventoty, for example host_initiators shold be skipped in this example:
```
ERROR -- evm: [ManageIQ::Providers::BaseManager::Refresher::PartialRefreshError]: PG::UndefinedTable: ERROR:  missing FROM-clause entry for table "physical_storages"
LINE 1: ...tors" WHERE "host_initiators"."ems_id" = $1 AND (("physical_...
```

Overriding the initialize_inventory_collections function, and manually removing empty iventory works but you said it should automaticly skip, so I'm looking on it. 

hopefully I will push a fix for both issues by tommorow.

agrare:
So that error means there is something wrong with the declaration of your association, best to push some specs so we can see the error and work through it.

agrare:
Example of inventory collections being skipped:
```
[----] D, [2021-12-01T10:13:39.152121 #906635:8d18] DEBUG -- evm: Skipping InventoryCollection:<StorageService>, strategy: local_db_find_missing_references processing because it will do no operation.
[----] D, [2021-12-01T10:13:39.152259 #906635:8d18] DEBUG -- evm: Skipping InventoryCollection:<ManageIQ::Providers::Autosde::StorageManager::CloudVolume>, strategy: local_db_find_missing_references processing because it will do no operation.
```

So that is working properly, the issue is the definition of the HostInitiator association/collection:
```
[----] D, [2021-12-01T10:13:39.152599 #906635:8d18] DEBUG -- evm: ----- BEGIN ----- Saving collection InventoryCollection:<ManageIQ::Providers::Autosde::StorageManager::HostInitiator>, strategy: local_db_find_missing_references of size 0 to the database, for the manager: 'ems_0000000000001'...
[----] D, [2021-12-01T10:13:39.154503 #906635:8d18] DEBUG -- evm: Processing InventoryCollection:<ManageIQ::Providers::Autosde::StorageManager::HostInitiator>, strategy: local_db_find_missing_references of size 0...
[----] E, [2021-12-01T10:13:39.156090 #906635:8d18] ERROR -- evm: Error when saving InventoryCollection:<ManageIQ::Providers::Autosde::StorageManager::HostInitiator>, strategy: local_db_find_missing_references with strategy: local_db_find_missing_references, saver_strategy: default, targeted: true. Message: PG::UndefinedTable: ERROR:  missing FROM-clause entry for table "physical_storages"
LINE 1: ...tors" WHERE "host_initiators"."ems_id" = $1 AND (("physical_...

     ActiveRecord::StatementInvalid:
       PG::UndefinedTable: ERROR:  missing FROM-clause entry for table "physical_storages"
       LINE 1: ...tors" WHERE "host_initiators"."ems_id" = $1 AND (("physical_...

```

Issue #108, opened 2021-11-23
Host Initiator Groups -- Creating New Host Initiator Groups


gregoryb-at-ibm:
Continuation of https://github.com/ManageIQ/manageiq-providers-autosde/issues/93

Part of https://github.com/ManageIQ/manageiq-providers-autosde/issues/83

In this issue we will allow users to create new host-initiator-groups through MIQ.

 - [x] https://github.com/ManageIQ/manageiq/pull/21572
 - [x] https://github.com/ManageIQ/manageiq-ui-classic/pull/7974
 - [x] https://github.com/ManageIQ/manageiq-api/pull/1105
 - [x] https://github.com/ManageIQ/manageiq-providers-autosde/pull/109
 - [x] https://github.com/ManageIQ/manageiq-schema/pull/619

PR #109, opened 2021-11-23
Creation of host initiator group

gregoryb-at-ibm:
In this set of PRs we will allow users to create new host-initiator-groups on their storage systems through MIQ.

Needs to be done after https://github.com/ManageIQ/manageiq-schema/pull/619

Part of https://github.com/ManageIQ/manageiq-providers-autosde/issues/108

image

image

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/host_initiator_group.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/host_initiator_group.rb
@@ -1,0 +1,12 @@
+class ManageIQ::Providers::Autosde::StorageManager::HostInitiatorGroup < ::HostInitiatorGroup
+  supports :create
+
+  def self.raw_create_host_initiator_group(ext_management_system, options = {})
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/autosde/storage_manager/host_initiator_group.rb
```diff
@@ -0,0 +1,12 @@
+class ManageIQ::Providers::Autosde::StorageManager::HostInitiatorGroup < ::HostInitiatorGroup
```
agrare, line 1: @gregoryb-at-ibm I just noticed that the `host_initiator_groups` model doesn't have a `:type` column so the subclass won't be set by the persister.

Can you open a PR to manageiq-schema to add a string type column for STI?

gregoryb-at-ibm: Will do. It's unrelated to this set of PRs though, right?

agrare (approved):

PR #111, opened 2021-11-30
Move boilerplate persister definitions to core

agrare:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/21579

### DIFF
--- a/app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
+++ b/app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
@@ -128,6 +128,8 @@ class ManageIQ::Providers::Autosde::Inventory::Parser::StorageManager < ManageIQ
         :name             => volume.name,
         :size             => volume.size * 1024 * 1024 * 1024,
         :ems_ref          => volume.uuid,
+        :volume_type      => "ISCSI/FC",
+        :bootable         => false,
         :storage_resource => persister.storage_resources.lazy_find(volume.storage_resource),
         :storage_service  => persister.storage_services.lazy_find(volume.service),
         :status           => volume.component_state,
--- a/app/models/manageiq/providers/autosde/inventory/persister/definitions/cloud_collections.rb
+++ b/app/models/manageiq/providers/autosde/inventory/persister/definitions/cloud_collections.rb
@@ -1,9 +1,0 @@
-module ManageIQ::Providers::Autosde::Inventory::Persister::Definitions::CloudCollections
-  extend ActiveSupport::Concern
-
-  def initialize_cloud_inventory_collections
-    %i[vms].each do |name|
-      add_collection(cloud, name)
-    end
-  end
-end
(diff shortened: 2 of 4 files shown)


agrare:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq/pull/21579

app/models/manageiq/providers/autosde/inventory/persister/storage_manager.rb
```diff
@@ -18,34 +18,11 @@ def initialize_inventory_collections
...
-  end
-
-  def shared_options
-    {
-      :strategy => strategy,
-      :targeted => targeted?,
-      :parent   => parent
-    }
-  end
-
```
agrare, line 36: NOTE all of these are defined by the base persister class already

app/models/manageiq/providers/autosde/storage_manager.rb
```diff
@@ -1,7 +1,14 @@
 class ManageIQ::Providers::Autosde::StorageManager < ManageIQ::Providers::StorageManager
+  require_nested :AutosdeClient
+  require_nested :CloudVolume
+  require_nested :ClusterVolumeMapping
+  require_nested :HostInitiatorGroup
+  require_nested :HostInitiator
+  require_nested :HostVolumeMapping
+  require_nested :PhysicalStorage
```
agrare, line 8: Without proper require_nested these constants won't be found automatically and the auto-model-class from the core persister won't work

app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
```diff
@@ -128,6 +128,8 @@ def cloud_volumes
         :name             => volume.name,
         :size             => volume.size * 1024 * 1024 * 1024,
         :ems_ref          => volume.uuid,
+        :volume_type      => "ISCSI/FC",
+        :bootable         => false,
```
agrare, line 132: Moved these out of the persister definition since they really belong in the parser

app/models/manageiq/providers/autosde/inventory/persister/definitions/cloud_collections.rb
```diff
@@ -1,9 +0,0 @@
-module ManageIQ::Providers::Autosde::Inventory::Persister::Definitions::CloudCollections
-  extend ActiveSupport::Concern
-
-  def initialize_cloud_inventory_collections
-    %i[vms].each do |name|
-      add_collection(cloud, name)
-    end
-  end
-end
```
agrare, line 9: This was never removed from the provider generated code

agrare:
@Fryguy this is green, PTAL

app/models/manageiq/providers/autosde/storage_manager.rb
agrare: I followed this up with some better specs to make sure this is working properly: https://github.com/ManageIQ/manageiq-providers-autosde/pull/112

PR #112, opened 2021-11-30
Improve AutoSDE Refresher Specs

agrare:
Depends on:
- [x] https://github.com/ManageIQ/manageiq-providers-autosde/pull/111


Fryguy:
Merged #111 - Can you rebase, cause this is hard to follow otherwise?

app/models/manageiq/providers/autosde/inventory/persister/storage_manager.rb
```diff
@@ -6,67 +6,15 @@ class ManageIQ::Providers::Autosde::Inventory::Persister::StorageManager < Manag
...
-    add_san_addresses
-    add_host_initiators
-    add_volume_mappings
-    add_wwpn_candidates
-    add_host_initiator_groups
-    add_physical_storages
-    add_physical_storage_families
-    add_storage_resources
-    add_storage_services
-    add_cloud_volumes
```
agrare, line 18: These were added so the subclass could include or exclude collections from the targeted persister.

This isn't necessary since empty targeted collections are skipped completely

agrare:
NOTE I split up the `git mv` of the inventory_spec -> refresher_spec into its own commit to make the changes there easier to follow

PR #114, opened 2021-12-01
Cleanup parse_targets! and the Targeted collector methods

agrare:


app/models/manageiq/providers/autosde/inventory/collector/target_collection.rb
```diff
@@ -18,15 +18,15 @@ def storage_resources
   end
 
   def storage_hosts
-    return [] if references(:storage_hosts).blank?
+    []
```
agrare, line 21: These were returning `nil` instead of `[]` if there were references to these (follow-up to https://github.com/ManageIQ/manageiq-providers-autosde/pull/107#discussion_r759270735)

app/models/manageiq/providers/autosde/inventory/collector/target_collection.rb
```diff
@@ -67,18 +67,13 @@ def parse_targets!
       case target
       when PhysicalStorage
         add_target(:physical_storages, target.ems_ref)
-        add_target(:physical_storage_families, PhysicalStorageFamily.find(target.physical_storage_family_id).ems_ref)
+        add_target(:physical_storage_families, target.physical_storage_family.ems_ref)
```
agrare, line 70: Can just use the association rather than an explicit query on the id cc @llivne

llivne: looks much better. 
you need to define physical_storage, the code should look like that:
```
      when CloudVolume
        physical_storage = target.storage_resource.physical_storage
        physical_storage_family = physical_storage.physical_storage_family

        add_target(:storage_resources, target.storage_resource.ems_ref)
        add_target(:storage_services, target.storage_service.ems_ref)
        add_target(:physical_storages, physical_storage.ems_ref)
        add_target(:physical_storage_families, physical_storage_family.ems_ref)
      end
```

agrare: :+1: good catch thanks

PR #115, opened 2021-12-01
Add targeted refresh specs

agrare:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/21582

TODO:
- [x] Test no targets
- [x] Test existing physical_storage target
- [x] Test new physical_storage 
- [x] Test existing cloud_volume target
- [x] Test new cloud_volume

### DIFF
--- a/app/models/manageiq/providers/autosde/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/autosde/inventory/collector/target_collection.rb
@@ -67,7 +67,6 @@ class ManageIQ::Providers::Autosde::Inventory::Collector::TargetCollection < Man
       case target
       when PhysicalStorage
         add_target(:physical_storages, target.ems_ref)
-        add_target(:physical_storage_families, target.physical_storage_family.ems_ref)
       when CloudVolume
         physical_storage = target.storage_resource.physical_storage
 
--- a/app/models/manageiq/providers/autosde/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/autosde/inventory/collector/target_collection.rb
@@ -68,13 +68,7 @@ class ManageIQ::Providers::Autosde::Inventory::Collector::TargetCollection < Man
       when PhysicalStorage
         add_target(:physical_storages, target.ems_ref)
       when CloudVolume
-        physical_storage = target.storage_resource.physical_storage
-
         add_target(:cloud_volumes, target.ems_ref)
-        add_target(:storage_resources, target.storage_resource.ems_ref)
-        add_target(:storage_services, target.storage_service.ems_ref)
-        add_target(:physical_storages, physical_storage.ems_ref)
-        add_target(:physical_storage_families, physical_storage.physical_storage_family.ems_ref)
       end
     end
   end
--- a/spec/models/manageiq/providers/autosde/storage_manager/refresher_spec.rb
+++ b/spec/models/manageiq/providers/autosde/storage_manager/refresher_spec.rb
@@ -129,6 +129,45 @@ describe ManageIQ::Providers::Autosde::StorageManager::Refresher do
         assert_inventory_not_changed { run_targeted_refresh(ems.cloud_volumes.find_by(:ems_ref => "ac287c2d-1776-48a3-a5c9-06327f4a57c4")) }
       end
 
+      it "with a new cloud volume" do
+        expect(volume_api)
+          .to receive(:volumes_get)
+          .and_return(
+            [
+              AutosdeOpenapiClient::VolumeResponse.new(
+                :compliant          => true,
+                :component_state    => "PENDING_CREATION",
+                :historical_service => nil,
+                :name               => "new-volume",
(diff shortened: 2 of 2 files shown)


app/models/manageiq/providers/autosde/inventory/collector/target_collection.rb
```diff
@@ -67,7 +67,6 @@ def parse_targets!
       case target
       when PhysicalStorage
         add_target(:physical_storages, target.ems_ref)
-        add_target(:physical_storage_families, PhysicalStorageFamily.find(target.physical_storage_family_id).ems_ref)
```
agrare, line 70: Since this is an existing record I don't know if we don't need to add a target for the physical_storage_family.
This would be required if
1. This were a new physical_storage and it is possible it has a new physical_storage_family that we haven't refreshed yet and thus the lazy find would fail
2. You want to capture changes to the physical_storage_family by refreshing the physical_storage, not sure why you wouldn't just target the physical_storage_family in that case though

spec/models/manageiq/providers/autosde/storage_manager/refresher_spec.rb
```diff
@@ -31,6 +30,64 @@
...
+      it "with no targets" do
+        assert_inventory_not_changed { run_targeted_refresh }
+      end
+
+      it "with an existing physical_storage target" do
+        expect(storage_system_api)
+          .to receive(:storage_systems_get)
+          .and_return(
+            [
+              AutosdeOpenapiClient::StorageSystem.new(
```
agrare, line 47: This is a duplicate of the payload that is returned for this storage system, the goal here is to run a targeted refresh that will result in 0 changes allowing us to `assert_inventory_not_changed`

We should have one of these for every targeted type that we support, this ensure that nothing is being deleted & recreated while targeting the different collections

spec/models/manageiq/providers/autosde/storage_manager/refresher_spec.rb
```diff
@@ -31,6 +30,64 @@
...
+
+        assert_inventory_not_changed { run_targeted_refresh(ems.physical_storages.first) }
+      end
+
+      def run_targeted_refresh(targets = [])
+        target = InventoryRefresh::TargetCollection.new(:manager => ems, :targets => Array(targets))
+
+        persister = ManageIQ::Providers::Autosde::Inventory::Persister::TargetCollection.new(ems)
+        collector = ManageIQ::Providers::Autosde::Inventory::Collector::TargetCollection.new(ems, target)
+        parser    = ManageIQ::Providers::Autosde::Inventory::Parser::StorageManager.new
```
agrare, line 74: NOTE I went this route instead of `described_class.refresh(target)` because this gives us more control over stubbing the autosde_client because we have access to the collector instance.

agrare:
@llivne this is the error you're talking about right? https://app.travis-ci.com/github/ManageIQ/manageiq-providers-autosde/jobs/550423532#L2125-L2142

llivne:
> @llivne this is the error you're talking about right? https://app.travis-ci.com/github/ManageIQ/manageiq-providers-autosde/jobs/550423532#L2125-L2142

@agrare yes this is the error. I cant find the reason for it, I suspect it has something to do with the associations (I think you also wrote about it) but still cant solve it. still working on it...

agrare:
@llivne check out https://github.com/ManageIQ/manageiq/pull/21582 I try do describe there why those aren't needed in this case

app/models/manageiq/providers/autosde/inventory/collector/target_collection.rb
llivne: Agree, this can be removed, we dont need this for now, the target refresh is only set for editing currently

app/models/manageiq/providers/autosde/inventory/collector/target_collection.rb
```diff
@@ -67,15 +67,8 @@ def parse_targets!
...
         add_target(:physical_storages, target.ems_ref)
-        add_target(:physical_storage_families, target.physical_storage_family.ems_ref)
       when CloudVolume
-        physical_storage = target.storage_resource.physical_storage
-
         add_target(:cloud_volumes, target.ems_ref)
-        add_target(:storage_resources, target.storage_resource.ems_ref)
-        add_target(:storage_services, target.storage_service.ems_ref)
-        add_target(:physical_storages, physical_storage.ems_ref)
-        add_target(:physical_storage_families, physical_storage.physical_storage_family.ems_ref)
```
agrare, line 78: You only need to add targets if your parser requires access to the data, so if your cloud_volume parser used `collector.storage_resources` directly you might need this.

agrare:
Okay we can test more collections but this is a good start, taking out of WIP

llivne (approved):

llivne:
@agrare its look fine by me. do you want me to merge, or you prefer someone else will take a look?

agrare:
:+1: thanks @llivne go ahead and merge it if it looks good to you

PR #119, opened 2021-12-07
Use targeted refresh for cloud_volume create/update/delete

agrare:


app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
```diff
@@ -18,8 +18,16 @@ def self.raw_create_volume(ext_management_system, options = {})
       :size    => options["size"],
       :count   => options["count"]
     )
-    ext_management_system.autosde_client.VolumeApi.volumes_post(vol_to_create)
-    EmsRefresh.queue_refresh(ext_management_system)
+
+    new_volume = ext_management_system.autosde_client.VolumeApi.volumes_post(vol_to_create)
```
agrare, line 22: Hey @llivne I am not able to confirm this, but the autosde_openapi_client gem docs show `volumes_post` returns a `VolumeResponse` same as `volumes_get` so I should be able to get the uuid of the new volume, can you check this when you get a chance?

llivne: that is correct, this is an example for a POST response:
```{
  "compliant": true,
  "component_state": "CREATED",
  "historical_service": 16,
  "name": "liran-idea",
  "service": "941f1e8c-4bf0-4d26-8e19-31942c24a451",
  "size": 1,
  "status": null,
  "storage_resource": "f85d4e2b-ab05-47b8-a568-fadac48d42cc",
  "unmapped_since": null,
  "uuid": "6a17f7e4-eff3-4d12-b58b-5e156e19067a"
}```

Comment (agrare) [reply app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb:22]: :+1: perfect thanks

**Review by llivne — APPROVED** [review]
I approve

**Comment by agrare** [comment]
Thanks for the review @llivne!  You can go ahead and merge if this looks good to you

File: app/models/manageiq/providers/autosde/inventory/collector/target_collection.rb [hunk 2]
```diff
@@ -1,8 +1,9 @@
 class ManageIQ::Providers::Autosde::Inventory::Collector::TargetCollection < ManageIQ::Providers::Autosde::Inventory::Collector::StorageManager
-  def initialize(_manager, _target)
+  def initialize(_manager, target)
     super
 
     parse_targets!
+    target.manager_refs_by_association_reset
```
agrare, line 6: NOTE this is needed because the `manager_refs_by_association` is cached when initially building the inventory collections, so later calls to `#add_target` don't update this.  I think this should be handled by the inventory_refresh gem when you call add_target so I'll work on adding that.

PR #123, opened 2021-12-20
Remove unused scopes from volume_mapping

agrare:
These were added originally to differentiate between a cluster and a host volume_mapping but it was (correctly) replaced by using STI.

These are now unused and can be deleted.

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/volume_mapping.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/volume_mapping.rb
@@ -2,12 +2,6 @@ class ManageIQ::Providers::Autosde::StorageManager::VolumeMapping < ::VolumeMapp
   HOST_MAPPING_OBJECT = 'host'.freeze
   HOST_GROUP_MAPPING_OBJECT = 'host_group'.freeze
 
-  scope :to_clusters,  -> { where("type LIKE ?", "%Cluster%") }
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/autosde/storage_manager/volume_mapping.rb
```diff
@@ -2,12 +2,6 @@ class ManageIQ::Providers::Autosde::StorageManager::VolumeMapping < ::VolumeMapp
   HOST_MAPPING_OBJECT = 'host'.freeze
   HOST_GROUP_MAPPING_OBJECT = 'host_group'.freeze
 
-  scope :to_clusters,  -> { where("type LIKE ?", "%Cluster%") }
-  scope :to_hosts,     -> { where("type LIKE ?", "%Host%") }
-
-  scope :to_clusters,  -> { where("type LIKE ?", "%Cluster%") }
-  scope :to_hosts,     -> { where("type LIKE ?", "%Host%") }
```
agrare, line 9: I'm actually very confused how these made it in, the first was from 30519c3b7d12eb9124297319aae902c6e843b679 which was included in https://github.com/ManageIQ/manageiq-providers-autosde/pull/102

The second was from commit 58b1785c9389f1021aacbe0d01a987ff7193832a but that doesn't appear to be part of a PR?

llivne (approved):

PR #103, opened 2021-10-26
[MORPHY] set minor version for autosde_openapi_client gem

llivne:
we started using version contorl for autosde_openapi_client gem.
the latest/master autosde-provider will be using version 1.1.x. older versions should use version 1.0.x.

we need to also backport this PR to lasker.

note
-----------
in the future we do want to set morphy to work with autosde_openapi_client gem version 1.1.x. in order to do it the following pr need to be backport to morphy:
https://github.com/ManageIQ/manageiq-providers-autosde/pull/99

PR #99, opened 2021-10-16
changing system-type in vcr cassete to the new format FlashedSystems/svc

llivne:
Physical-storage system_type was changesd in autosde from svc to FlashSystems/SVC. miq dynamicly imports the names of the system-types from autosde so no changes are needed in miq code, except for the rspec vcr-cassetes that save the nave svc which is no longer supported. 
vcr where modified to FlashSystems/SVC

PR #127, opened 2022-01-11
[MORPHY] set autosde openapi_client_gem version to 1.1.0

llivne:
Physical-storage system_type was changesd in autosde from svc to FlashSystems/SVC.
This should also be part of the M release

links
-------
related info about this PR can be found in:
https://github.com/ManageIQ/manageiq-providers-autosde/pull/99
https://github.com/ManageIQ/manageiq-providers-autosde/pull/103


agrare (approved):

agrare:
For background, we didn't originally include this enhancement in morphy because the gem introduced a breaking API change that caused the release branches to fail.

AutoSDE team wants this feature backported to the morphy branch so we're going to bump the gem version here

agrare:
Closing in favor of backporting https://github.com/ManageIQ/manageiq-providers-autosde/pull/99

Fryguy:
Using this as the backport PR for #99

PR #129, opened 2022-01-25
Override display name for volume_mappings in subclasses

agrare:
Proposed alternative to https://github.com/ManageIQ/manageiq/pull/21640

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/cluster_volume_mapping.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/cluster_volume_mapping.rb
@@ -1,6 +1,5 @@
-class ManageIQ::Providers::Autosde::StorageManager::ClusterVolumeMapping <
-  ManageIQ::Providers::Autosde::StorageManager::VolumeMapping
-  def mapped_to
-    host_initiator_group
+class ManageIQ::Providers::Autosde::StorageManager::ClusterVolumeMapping < ManageIQ::Providers::Autosde::StorageManager::VolumeMapping
+  def name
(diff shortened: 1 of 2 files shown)


Fryguy:
Grammar-wise the "name" being "is mapped to" is strange, though I guess it's ok for now as it's being presented as "Details" in the UI.  We may want something other than "name" to be the thing we override, perhaps "details" or "display_name".

agrare:
Yeah details or description would both work, there is an open UI PR but I didn't see name being called out specifically for a volume_mapping...name might just be the default for the textual_summary

llivne:
@agrare I'm replacing greg on this issue. I tested your proposed alternative and its working like expected. I'm ok with it, and I also saw that @Fryguy reverted https://github.com/ManageIQ/manageiq/pull/21640 so this PR can be merged. 

Should I do it, or you prefer someone from miq team ?

llivne (approved):

agrare:
@llivne since https://github.com/ManageIQ/manageiq-ui-classic/pull/8007 was already merged you can go ahead and merge this one, can you follow-up that UI PR to work with a different attribute name?  Details or description per @Fryguy's comment above?

llivne:
> @llivne since [ManageIQ/manageiq-ui-classic#8007](https://github.com/ManageIQ/manageiq-ui-classic/pull/8007) was already merged you can go ahead and merge this one, can you follow-up that UI PR to work with a different attribute name? Details or description per @Fryguy's comment above?

@agrare @Fryguy like I said I'm replacing greg on this issue, since greg left IBM. I took another look and I agree that the description is a bit strange. I tried to find a way to solve it in this PR, but I need to do also changes in the ui repo, and also I prefer to have additional feedback on how to do the changes from our team and architect so for now I merged this PR and will open an issue to change the attribute name

PR #142, opened 2022-03-15
set storage resources size unit to GB instead of bytes

llivne:
Description
------
When we click on Storage->Storage Resources the list of storage resources shows the Total Size and Free Space in Bytes. From a User experience perspective, displaying them in TB or GB would be ideal as the storage allocated will be always larger in size and showing them in bytes does not help

Acceptance Criteria
-----
The Total Size and Free Space values in the list must be displayed in GB's or TB's instead of Bytes 

Business Justification
-----
User experience

before
-----

after
-----

links
-----
https://github.com/ManageIQ/manageiq-ui-classic/pull/8176


llivne:
@Fryguy I'm closing this PR and will do the modifications in the ui-classic like you suggested

PR #143, opened 2022-03-28
get error data and return error response error

eilam20:
Change the autosde response error from json message to detail string message.
Before:
Screen Shot 2022-03-24 at 15 33 09

After
Screen Shot 2022-03-28 at 16 36 08

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
@@ -70,7 +70,9 @@ class ManageIQ::Providers::Autosde::StorageManager::AutosdeClient < AutosdeOpena
     data = self.AuthenticationApi.token_auth_post(auth_request, opts)
     @token = data.token
   rescue => e
-    raise AUTH_ERRR_MSG + e.to_s
+    error_data = JSON.parse(e.response_body)
+    raise AUTH_ERRR_MSG + error_data['detail']['non_field_errors'][0]
+
   end
 
   private
--- a/app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
@@ -69,10 +69,11 @@ class ManageIQ::Providers::Autosde::StorageManager::AutosdeClient < AutosdeOpena
     opts = {:login => true}
     data = self.AuthenticationApi.token_auth_post(auth_request, opts)
     @token = data.token
-  rescue => e
+  rescue AutosdeOpenapiClient::ApiError => e
     error_data = JSON.parse(e.response_body)
     raise AUTH_ERRR_MSG + error_data['detail']['non_field_errors'][0]
-
+  rescue => e
+    raise AUTH_ERRR_MSG + e.to_s
   end
 
   private


app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
```diff
@@ -70,7 +70,9 @@ def login
     data = self.AuthenticationApi.token_auth_post(auth_request, opts)
     @token = data.token
   rescue => e
-    raise AUTH_ERRR_MSG + e.to_s
+    error_data = JSON.parse(e.response_body)
```
agrare, line 73: Thanks for the PR @eilam20 !
Only issue I see here is that this `rescue` block will catch any exception type, and while `e.to_s` will work with any type of exception `e.response_body` will not.

Can you catch a more specific type of exception like `rescue AutosdeOpenapiClient::ApiError => e` for parsing the response_body, then leave this "fallback" exception handler the way it is?

eilam20:
Thanks for the review, I fixed the code as suggested .

llivne:
@eilam20 the RSpecs fails. the source is the new autosde_client gem that was just published. please Fix the RSPEC in a seperate PR before merging this PR.

llivne:
@eilam20 I fixed the rspec in https://github.com/ManageIQ/manageiq-providers-autosde/pull/144. please wait for it to be merged and then rebase. Make sure all the tests pass in this PR

app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
```diff
@@ -69,6 +69,9 @@ def login
     opts = {:login => true}
     data = self.AuthenticationApi.token_auth_post(auth_request, opts)
     @token = data.token
+  rescue AutosdeOpenapiClient::ApiError => e
+    error_data = JSON.parse(e.response_body)
+    raise AUTH_ERRR_MSG + error_data['detail']['non_field_errors'][0]
```
agrare, line 74: We could code this a little more defensively just in case `detail` and/or `non_field_errors` are empty, e.g.:
```ruby
raise AUTH_ERRR_MSG + error_data.dig("detail", "non_field_errors")&.first
```

But this can be tackled in a follow-up.

agrare: Also, was `AUTH_ERRR_MSG` intentional or should it be `AUTH_ERROR_MSG`?

agrare: And as a project MIQ prefers string interpolation (aka `"#{AUTH_ERRR_MSG}#{error_data.dig("detail", "non_field_errors")&.first}"`) over string addition like you have here.  I think rubocop didn't warn about it because it hasn't picked up on the fact that `AUTH_ERRR_MSG` is a string

Fryguy:
Backported to `najdorf` in commit 3a6fc5c843d8749b9214c794cc6161c83bb05569.

```text
commit 3a6fc5c843d8749b9214c794cc6161c83bb05569
Author: Adam Grare <adam@grare.com>
Date:   Tue Mar 29 15:38:25 2022 -0400

    Merge pull request #143 from Autosde/change_autosde_respose_error
    
    get error data and return error response error
    
    (cherry picked from commit 8a9ce5d6da4135b991c948f0a7658954cc7410e5)
```

PR #149, opened 2022-05-25
implemented validation storage system button

TonySchneider:
# Description
We implemented a 'validate-storage' rest API on the AutoSDE side to have the ability to validate the storage system credentials before we send the creation task to the queue. With this feature, we're saving time and can change the credentials (if there are provided some invalid parameters) without waiting for the response of the queued task.
I implemented the validate button on the attaching storage system page. So we can press the button and validate that the provided parameters are valid. If they are valid, AutoSDE saved this validation request and when we'll press the submit button, the queue task will not validate another time but just take the object that has been created by the previous validation request.

# Illustration video
validate-button

# Links
https://github.com/ManageIQ/manageiq/pull/21882
https://github.com/ManageIQ/manageiq-ui-classic/pull/8279
https://github.com/ManageIQ/manageiq-api/pull/1163

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
@@ -14,23 +14,13 @@ class ManageIQ::Providers::Autosde::StorageManager::PhysicalStorage < ::Physical
   end
 
   def self.raw_validate_physical_storage(ext_management_system, options = {})
-    header_params = options[:header_params] || {}
-    header_params['Accept'] = ext_management_system.autosde_client.select_header_accept(['*/*'])
-    header_params['Content-Type'] = ext_management_system.autosde_client.select_header_content_type(['application/json'])
-
-    request_opts = {
-      :header_params => header_params,
-      :body => options.merge(
-        :system_type => PhysicalStorageFamily.find(options['physical_storage_family_id']).name,
-      ),
-      :auth_names => ['bearerAuth'],
-    }
-
-    data, status_code, headers = ext_management_system.autosde_client.call_api(:POST, 'validate-system', request_opts)
-    if ext_management_system.autosde_client.config.debugging
-      ext_management_system.autosde_client.config.logger.debug "API called: StorageSystemValidation#storage_systems_post\nData: #{data.inspect}\nStatus code: #{status_code}\nHeaders: #{headers}"
-    end
-    return data, status_code, headers
+    validation_object = ext_management_system.autosde_client.StorageSystemCreate(
+      :management_ip => options['management_ip'],
+      :user          => options['user'],
+      :password      => options['password'],
+      :system_type   => PhysicalStorageFamily.find(options['physical_storage_family_id']).name
+    )
+    ext_management_system.autosde_client.ValidateSystemApi.validate_system_post(validation_object)
   end
 
   def raw_update_physical_storage(options = {})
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
```diff
@@ -13,6 +13,26 @@ def raw_delete_physical_storage
...
+
+    request_opts = {
+      :header_params => header_params,
+      :body => options.merge(
+        :system_type => PhysicalStorageFamily.find(options['physical_storage_family_id']).name,
+      ),
+      :auth_names => ['bearerAuth'],
+    }
+
+    data, status_code, headers = ext_management_system.autosde_client.call_api(:POST, 'validate-system', request_opts)
```
agrare, line 29: Looks like an API call which should be added to the gem so you can call it without having to set up the headers/etc... yourself?

app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
```diff
@@ -13,6 +13,26 @@ def raw_delete_physical_storage
...
 
+  def self.raw_validate_physical_storage(ext_management_system, options = {})
+    header_params = options[:header_params] || {}
+    header_params['Accept'] = ext_management_system.autosde_client.select_header_accept(['*/*'])
+    header_params['Content-Type'] = ext_management_system.autosde_client.select_header_content_type(['application/json'])
+
+    request_opts = {
+      :header_params => header_params,
+      :body => options.merge(
+        :system_type => PhysicalStorageFamily.find(options['physical_storage_family_id']).name,
```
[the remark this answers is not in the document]
TonySchneider: agree even though the backend takes only needed info. but you're right that we shouldn't send garbage. I'll fix it.

app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
TonySchneider: The objects that gem uses are created by the jssonet file (generation tool) but the 'validate-credentials' wasn't implemented like that but as a static endpoint (for now). I can't use the gem function. Even though the endpoint in AutoSDE was implemented by using the generation tool, there is a function called 'storage_systems_post_with_http_info' (path: autosde_openapi_client/api/storage_system_api.rb) which can fit this API call but this function gets only StorageSystemCreate object, but I need to send different object. StorageSystemValidate for example which we didn't use the generation feature for now (as I already wrote). So, the only approach to this API call is to work with the function without using an object.

We will change the endpoint in the future to be generated by the jssonet file and then the gem can use the StorageSystemValidate object but then I'll need to create a new function as well so I can use this object instead of StorageSystemCreate.

TonySchneider:
Hey, any new updates?

app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
TonySchneider: @agrare I've changed the implementation of the validation endpoint to be generated by the jsonnet tool (in AutoSDE). I hope we'll push it today to the master and I will use the gem function as you suggested. So the "article" that I wrote above doesn't relevant anymore.

TonySchneider:
I've fixed Rubocop syntax warnings

agrare:
Hey @TonySchneider I'd like to wait for this method to be implemented in the gem rather than hack in the API call in this way

TonySchneider:
Hey @agrare, I've changed the api call to gem function.

PR #151, opened 2022-06-22
EventCatcher worker for Physical Storage

TonySchneider:
We decided to import critical alerts from the physical storage into MIQ, catch them by a new listener worker, store them in the queue and convert them to EventStream objects so we'll be able to define Physical Storage Alerts and show them on the monitor page.

I've already implemented the EventCatcher, and Stream and configured the necessary files accordingly.
This worker should work as a service and pull our events from AutoSDE every 15 seconds.

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/event_catcher.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/event_catcher.rb
@@ -1,0 +1,4 @@
+class ManageIQ::Providers::Autosde::StorageManager::EventCatcher < ManageIQ::Providers::BaseManager::EventCatcher
+  require_nested :Runner
+  require_nested :Stream
+end
--- a/app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
@@ -1,0 +1,45 @@
+class ManageIQ::Providers::Autosde::StorageManager::EventCatcher::Runner < ManageIQ::Providers::BaseManager::EventCatcher::Runner
+  def event_monitor_handle
+    @event_monitor_handle ||= self.class.parent::Stream.new(@ems)
+  end
+
+  def reset_event_monitor_handle
+    @autosde_client = nil
+  end
+
+  def stop_event_monitor
+    event_monitor_handle.stop
+  end
+
+  def monitor_events
+    event_monitor_handle.start
+
+    event_monitor_running
+
+    event_monitor_handle.poll_events do |event|
+      @queue.enq(event)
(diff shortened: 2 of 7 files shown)


TonySchneider:
Hi @agrare, I have implemented the EventCatcher. I looked at the VMWare implementation and this guide:
https://www.manageiq.org/docs/guides/providers/events and it works.
It indeed pulls events from AutoSDE and converts them to EventStreamer objects but I am stuck on the second part - present them on the monitor alerts page. I tried to understand what the VMWare team did but they have many Alerts types which I can't take and use for the physical storage. basically, I need to show these event streams object on the monitor and I can't figure out how to.
Maybe you or one of the MIQ guys can assist me with it? or you are familiar with some tutorials that can help me.

app/models/manageiq/providers/autosde/storage_manager/event_catcher/stream.rb
```diff
@@ -0,0 +1,29 @@
+class ManageIQ::Providers::Autosde::StorageManager::EventCatcher::Stream
+  attr_reader :ems, :stop_polling, :poll_sleep
+
+  def initialize(options = {})
+    @ems = ManageIQ::Providers::Autosde::StorageManager.first
```
agrare, line 5: This won't work if there is more than one instance of an Autosde provider.  Recommend passing in the ems as an arg:
```suggestion
  def initialize(ems, options = {})
    @ems = ems
```

Then when you initialize the Stream class in your Runner simply pass in the ems ivar:
`@event_monitor_handle ||= self.class.parent::Stream.new(@ems)`

app/models/manageiq/providers/autosde/storage_manager/event_catcher.rb
```diff
@@ -0,0 +1,3 @@
+class ManageIQ::Providers::Autosde::StorageManager::EventCatcher < ManageIQ::Providers::BaseManager::EventCatcher
+  require_nested :Runner
```
agrare, line 2: @TonySchneider you also need to `require_nested :Stream` here

app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  private
+
+  def event_to_hash(event, ems_id)
+    {
+      :event_type => event.event_type,
+      :source     => "AUTOSDE",
+      :ems_ref    => event.event_id,
+      :timestamp  => event.last_timestamp,
+      :full_data  => event,
```
agrare, line 40: What type of object is this?  We need to ensure this is a hash not an object

PR #154, opened 2022-07-20
display events for physical storage

TonySchneider:
* Following this PR: https://github.com/ManageIQ/manageiq-providers-autosde/pull/151
This is an implementation of the monitoring page for physical storage where we'll show our critical alerts (storage events).

Before:
image

After:
image

# related PRs:
- https://github.com/ManageIQ/manageiq-schema/pull/656
- https://github.com/ManageIQ/manageiq/pull/22008
- https://github.com/ManageIQ/manageiq-ui-classic/pull/8373

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
@@ -33,13 +33,14 @@ class ManageIQ::Providers::Autosde::StorageManager::EventCatcher::Runner < Manag
 
   def event_to_hash(event, ems_id)
     {
-      :event_type => event.event_type,
-      :source     => "AUTOSDE",
-      :ems_ref    => event.event_id,
-      :timestamp  => event.last_timestamp,
-      :full_data  => event.to_hash,
-      :ems_id     => ems_id,
-      :message    => event.description
+      :event_type               => event.event_type,
+      :source                   => "AUTOSDE",
+      :ems_ref                  => event.event_id,
+      :physical_storage_ems_ref => event.storage_system,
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
```diff
@@ -33,13 +33,14 @@ def queue_event(event)
...
-      :source     => "AUTOSDE",
-      :ems_ref    => event.event_id,
-      :timestamp  => event.last_timestamp,
-      :full_data  => event.to_hash,
-      :ems_id     => ems_id,
-      :message    => event.description
+      :event_type               => event.event_type,
+      :source                   => "AUTOSDE",
+      :ems_ref                  => event.event_id,
+      :physical_storage_ems_ref => event.storage_system,
```
agrare, line 39: NOTE as we handle events for more types of inventory this might be an issue unless `event.storage_system` is always present.

agrare: e.g. what a lot of providers do is build the "base" event hash, then depending on the event type add in keys for the related inventory but this is something we can get to later

agrare (approved):

app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
TonySchneider: storage_system can't be null in AutoSDE. it's a mandatory field.
As StorageResources, Volumes, etc...

agrare (approved):

PR #156, opened 2022-07-26
fixed edit cloud not working

Oferlis:
someone changed edit_resource a few months ago inside `manageiq\plugins\manageiq-providers-autosde\app\models\manageiq\providers\autosde\storage_manager\physical_storage.rb` and now we pass symbols instead of keys.

We modified the `raw_update_volume` from `@name => options["name"]` to `@name => options[:name]` in order for it to work.


agrare (approved):
This is a good fix, but generally we've been switching to string keys for these operations because that is what the API naturally returns from the JSON payload.  Cloud Volume edit_resource happens to have deep_symbolize_keys before passing the payload back down but that is not the norm.

Thanks @Oferlis !

PR #160, opened 2022-08-07
Added support for multiple IQNs, parsing the array and passing it

Oferlis:
Added parsing and organizing multiple IQNs from array to string to be sent to AutoSDE for addition.

Another PR in manageiq_ui_classic will be opened to reflect the UI change:
https://github.com/ManageIQ/manageiq-ui-classic/pull/8398

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
@@ -4,7 +4,6 @@ class ManageIQ::Providers::Autosde::StorageManager::HostInitiator < ::HostInitia
 
   def self.raw_create_host_initiator(ext_management_system, options = {})
     # WWPN/IQN values sent to autosde should be format as colon separated string (e.g. WWPN1:WWPN2:WWPN3)
-    
     wwpns = Array(options['custom_wwpn'])
     wwpns += Array(options['wwpn']).map { |item| item["value"] }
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
```diff
@@ -5,15 +5,17 @@ class ManageIQ::Providers::Autosde::StorageManager::HostInitiator < ::HostInitia
   def self.raw_create_host_initiator(ext_management_system, options = {})
     wwpns = Array(options['custom_wwpn'])
     wwpns += Array(options['wwpn']).map { |item| item["value"] }
-
-    # wwpn values send to autosde should be format as colon separated string (e.g. WWPN1:WWPN2:WWPN3)
```
agrare, line 9: Maybe better to move this comment up above `wwpns` here https://github.com/ManageIQ/manageiq-providers-autosde/pull/160/files#diff-41fa59bd46caf3a422d2250625c4a970d5a0eb9840ba37722b2a73b1407b171aR6 since it applies to wwpns and iqns

Also, should be `wwpn values sent` not `send` to autosde

Oferlis: I modified the comments as you suggested.

agrare (approved):

PR #161, opened 2022-08-09
Sde 737 add autosde catalog auto

eilam20:
create a volume dialog, autosde catalog and autosde catalog item after adding a physical storage.

related to:
https://github.com/ManageIQ/manageiq-content/pull/715


app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
```diff
@@ -17,18 +17,18 @@ def raw_delete_physical_storage
   def self.raw_validate_physical_storage(ext_management_system, options = {})
     validation_object = ext_management_system.autosde_client.StorageSystemCreate(
       :management_ip => options['management_ip'],
-      :user          => options['user'],
-      :password      => options['password'],
-      :system_type   => PhysicalStorageFamily.find(options['physical_storage_family_id']).name
+      :user => options['user'],
+      :password => options['password'],
+      :system_type => PhysicalStorageFamily.find(options['physical_storage_family_id']).name
```
agrare, line 22: We still want to keep the `=>` lined up like they were before

app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
```diff
@@ -37,21 +37,31 @@ def raw_update_physical_storage(options = {})
...
+      ### SETUP SERVICE CATALOG COMPONENTS ###
+      # CATALOG
+      catalog_id = ServiceTemplateCatalog.find_or_create_by(:name => "AutoSDE Catalog").id
+      # DIALOG
+      vol_dialog = YAML.load_file('plugins/manageiq-providers-autosde/service_catalog/volume/dialog.yaml')
+      vol_dialog_id = DialogImportService.new.import(vol_dialog).id
+      # ITEM
+      vol_item_prop = { :name => "Volume", :prov_type => "autosde", :display => "true", :service_template_catalog_id => catalog_id,
+                        :config_info => { :provision => { :fqname => "/Storage/AutoSde/Services/ServiceVolumeRequestApproval/Default", :dialog_id => vol_dialog_id } } }
+      ServiceTemplate.create_catalog_item(vol_item_prop, nil) unless ServiceTemplate.find_by(:name => vol_item_prop[:name]) != nil
```
agrare, line 58: Is this just for your dev/testing?  If not then I don't think this is the right place to do this

eilam20: We want to create for autosde clients a service catalog item that will automatically create a volume with the appropriate dialog.
The catalog will be available after connecting to physical storage.
Where should we put this script? in another function? or another file?

eilam20: I put it in another file.

eilam20:
@agrare can you review it?

app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
agrare: Setting up a service catalog should be done by the user not automatically in code like this.

content/miq/dialogs/volume.yaml
```diff
@@ -0,0 +1,247 @@
+---
```
agrare, line 1: If you put this in `content/miq_dialogs` it will be automatically seeded for you.  Also since these are seeded into the global namespace please include the operation and provider name like `cloud_volume_autosde_order.yml`

eilam20: where the dialog will be automatically seeded?

content/miq_dialogs/cloud_volume_autosde_order.yaml
```diff
@@ -0,0 +1,247 @@
+---
+description: Order Storage Volume
+buttons: submit,cancel
```
agrare, line 3: This is failing:
```
ActiveModel::UnknownAttributeError: unknown attribute 'buttons' for MiqDialog.
/home/grare/adam/src/manageiq/manageiq/app/models/miq_dialog/seeding.rb:49:in `seed_record'
/home/grare/adam/src/manageiq/manageiq/app/models/miq_dialog/seeding.rb:13:in `block (2 levels) in seed'
/home/grare/adam/src/manageiq/manageiq/app/models/miq_dialog/seeding.rb:12:in `each'
/home/grare/adam/src/manageiq/manageiq/app/models/miq_dialog/seeding.rb:12:in `block in seed'
/home/grare/adam/src/manageiq/manageiq/app/models/miq_dialog/seeding.rb:9:in `seed'
```

I think what you want is:
```
:content:
  :buttons:
  - :submit
  - :cancel
```

like https://github.com/ManageIQ/manageiq/blob/master/product/dialogs/miq_dialogs/miq_provision_dialogs.yaml#L5-L8

content/miq/dialogs/volume.yaml
agrare: Dialogs are seeded here https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_dialog/seeding.rb this will automatically create MiqDialog database records that the user can select

content/miq_dialogs/cloud_volume_autosde_order.yaml
agrare: If I apply this patch it seeds:
```
adam@desktop:~/src/manageiq/manageiq-providers-autosde$ git diff
diff --git a/content/miq_dialogs/cloud_volume_autosde_order.yaml b/content/miq_dialogs/cloud_volume_autosde_order.yaml
index 0c0bfae..21847cb 100644
--- a/content/miq_dialogs/cloud_volume_autosde_order.yaml
+++ b/content/miq_dialogs/cloud_volume_autosde_order.yaml
@@ -1,9 +1,11 @@
 ---
+name: cloud_volume_autosde_order
 description: Order Storage Volume
-buttons: submit,cancel
-label: AutoSDE Dialog
-system: false
-dialog_tabs:
+:content:
+  :buttons:
+  - :submit
+  - :cancel
+  :dialogs:
   - description:
     display:
     label: Volume
```

eilam20:
@agrare there is a directory where we can put our script for creating the dialog, catalog, and catalog item?
is this something that other providers did?

agrare:
@eilam20 creating catalogs and catalog items is something that users do since access and rules around the catalog are very user specific, see https://www.manageiq.org/docs/get-started/create-service-item.html for the steps there.

If you fix the dialog yaml then the dialog you created will show up as an option in the dialogs list.

content/miq_dialogs/cloud_volume_autosde_order.yaml
tzurvaza: image

Hi @agrare this yaml originally came from the **export dialog tool**, There is a better way to export the dialog?

eilam20:
@agrare we decided to close this pr, the users will create the catalog item manually or with an external script.

PR #162, opened 2022-08-17
change names of calls to StorageHostsMappingApi

Lauferin:
The name of a class and a model in the oas client has changed, so when calling to the old name, an error is raised.

StorageHostVolumeMappingApi -> StorageHostsMappingApi
StorageHostVolumeMapping -> StorageHostsMapping
StorageHostVolumeMappingCreate -> StorageHostsMappingCreate

I changed the calls to the api and models to fix it in the two files where it is relevant
errorMappingApiCall
.

### DIFF
--- a/app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
+++ b/app/models/manageiq/providers/autosde/inventory/collector/storage_manager.rb
@@ -16,7 +16,7 @@ class ManageIQ::Providers::Autosde::Inventory::Collector::StorageManager < Manag
   end
 
   def host_volume_mappings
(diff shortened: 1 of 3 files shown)


agrare:
@Lauferin what is the new gem version?  This is a breaking API change and the gem should have a major version bump

Lauferin:
@agrare We created a new gem and changed version to 1.2, everything should be fine now.

agrare:
Okay, we want to follow semantic versioning (https://semver.org/) so any breaking change should be a major version bump not a minor one as you have here.

I'm going to merge this but follow semver going forward

agrare:
@Lauferin you also need to yank the "bad" version (1.1.35 thru 1.1.40) as they are breaking release branch builds

Lauferin:
@agrare , yanked and taggedto 1.1.34.

PR #164, opened 2022-08-19
[WIP] Replace GIT with PATH in Gemfile.lock

agrare:
Prevent the Gemfile.lock from being rebuilt because of the change from a GIT reference to a PATH reference when bundling a plugin

When we use the `Gemfile.lock.release` from a core release branch then bundle install the `GIT` gem for the plugin currently being used is replaced by a `PATH` reference.  This unlocks the dependent gems and allows them to be updated which we don't want.

If we manually replace the specific plugin we are testing and leave the rest of the Gemfile.lock alone then the original gem versions are maintained.

VERY WIP

I used a little ruby script to do this for now, I believe I can get this to work with an awk state machine but it will take some time.
A state machine like `/GIT\n  remote: .+manageiq-providers-autosde/, /specs:/ {}` should allow us to print what we want changed between the start-pattern and end-pattern.

TODO:
- [ ] replace ruby with awk (?)
- [ ] parameterize the repo url

### DIFF
--- a/bin/before_install
+++ b/bin/before_install
@@ -18,3 +18,14 @@ elif [ ! -d "$spec_manageiq" ]; then
   echo "== Cloning manageiq sample app =="
   git clone https://github.com/ManageIQ/manageiq.git --branch master --depth 1 "$spec_manageiq"
 fi
+
+if [ -n "$CI" ]; then
+  # Gemfile.lock.release only applies to non-master branches and PRs to non-master branches
+  if [[ "$GITHUB_REPOSITORY_OWNER" = "ManageIQ" && "$GITHUB_BASE_REF" != "master" && "$GITHUB_REF_NAME" != "master" && "$GITHUB_REF_NAME" != "dependabot/"* ]]; then
+    echo "== Setup Gemfile.lock.release =="
+    cp -f "$spec_manageiq/Gemfile.lock.release" "$gem_root/Gemfile.lock"
+
+    $spec_manageiq/bin/ci/patch_plugin_gemfile_lock
+    echo
+  fi
+fi


bin/before_install
```diff
@@ -18,3 +18,13 @@ elif [ ! -d "$spec_manageiq" ]; then
   echo "== Cloning manageiq sample app =="
   git clone https://github.com/ManageIQ/manageiq.git --branch master --depth 1 "$spec_manageiq"
 fi
+
+if [ -n "$CI" ]; then
+  # Gemfile.lock.release only applies to non-master branches and PRs to non-master branches
+  if [[ "$GITHUB_REPOSITORY_OWNER" = "ManageIQ" && "$GITHUB_BASE_REF" != "master" && "$GITHUB_REF_NAME" != "master" && "$GITHUB_REF_NAME" != "dependabot/"* ]]; then
+    echo "== Setup Gemfile.lock.release =="
+    cp -f "$spec_manageiq/Gemfile.lock.release" "$gem_root/Gemfile.lock"
+    ruby -e 'File.write("Gemfile.lock", File.read("Gemfile.lock").gsub(/GIT\n  remote: #{Regexp.escape(ENV["GITHUB_SERVER_URL"])}\/#{Regexp.escape(ENV["GITHUB_REPOSITORY"])}\n  revision: .+\n  branch: .+\n/, "PATH\n  remote: .\n"))'
```
Fryguy, line 27: Unfortunately, Ruby isn't installed by this point in CI :(  But this has to be here before ruby is installed because it also installs the gems.

bin/before_install
```diff
@@ -18,3 +18,26 @@ elif [ ! -d "$spec_manageiq" ]; then
...
+f = open('Gemfile.lock')
+gemfile_lock = f.read()
+f.close()
+gemfile_lock_patched = re.compile('GIT\n  remote: https://github.com/ManageIQ/manageiq-providers-autosde\n  revision: .+\n  branch: .+\n').sub('PATH\n  remote: .\n', gemfile_lock)
+f = open('Gemfile.lock', 'w')
+f.write(gemfile_lock_patched)
+f.close()
+EOF
+)
+    python3 -c "$prog"
```
Fryguy, line 40: Once this goes green, I'm wondering if the "right" approach is to store this bit of the script in the core repo.  The reason is that at this point we now have the core repo cloned, and then we can start sharing from the core repo.  This way we don't copy-paste across all of the repos.

PR #166, opened 2022-08-23
Set the event field  of PhysicalStorage as the physical-storage name

tzurvaza:
This change will show on the event-monitor the source of  physical-storage as the name (instead the name of the provider), so when we open the event-monitor of Storage-Manager we can understand which physical-storage is reporting us.

# Before
image

# After
image

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
@@ -34,7 +34,7 @@ class ManageIQ::Providers::Autosde::StorageManager::EventCatcher::Runner < Manag
   def event_to_hash(event, ems_id)
     {
       :event_type               => event.event_type,
-      :source                   => "AUTOSDE",
+      :source                   => PhysicalStorage.find_by(:ems_ref=>event.storage_system).name,
       :ems_ref                  => event.event_id,
       :physical_storage_ems_ref => event.storage_system,
       :timestamp                => event.created_at,


llivne (approved):

ereztheodorou (approved):

app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
```diff
@@ -34,7 +34,7 @@ def queue_event(event)
   def event_to_hash(event, ems_id)
     {
       :event_type               => event.event_type,
-      :source                   => "AUTOSDE",
+      :source                   => PhysicalStorage.find_by(:ems_ref=>event.storage_system).name,
```
agrare, line 37: `:source` must be the provider type for the rest of the event switchboard to work so this will break all automate event handlers.

Also the event catcher should not perform any queries like this for a few reasons but two primary among them being performance and the fact that events might be for inventory which is not in the database yet.  This is filled in by the event_handler worker already (https://github.com/ManageIQ/manageiq/blob/master/app/models/ems_event.rb#L140-L142, https://github.com/ManageIQ/manageiq/blob/0a7431d0b1b52cca9b05839c055bc25b51dd9506/app/models/ems_event.rb#L106)

Also if the physical storage fails to be found here (e.g. if it hasn't been refreshed yet) then this `.find_by` will return `nil` and `nil.name` will throw an exception.

agrare (commented):
We need to revert this

app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
tzurvaza: @agrare I saw that there is `physical_storage_name`  in the event stream:
image
Can you give me direction how to display it in the monitor's UI, something like that: image

PR #168, opened 2022-08-24
[OPARIN] [WIP] Oparin replace git with path in gemfile lock

agrare:
DONT MERGE

### DIFF
--- a/bin/before_install
+++ b/bin/before_install
@@ -18,3 +18,14 @@ elif [ ! -d "$spec_manageiq" ]; then
   echo "== Cloning manageiq sample app =="
   git clone https://github.com/ManageIQ/manageiq.git --branch oparin --depth 1 "$spec_manageiq"
 fi
+
+if [ -n "$CI" ]; then
+  # Gemfile.lock.release only applies to non-master branches and PRs to non-master branches
+  if [[ "$GITHUB_REPOSITORY_OWNER" = "ManageIQ" && "$GITHUB_BASE_REF" != "master" && "$GITHUB_REF_NAME" != "master" && "$GITHUB_REF_NAME" != "dependabot/"* ]]; then
+    echo "== Setup Gemfile.lock.release =="
+    cp -f "$spec_manageiq/Gemfile.lock.release" "$gem_root/Gemfile.lock"
+
+    $spec_manageiq/bin/ci/patch_plugin_gemfile_lock
+    echo
+  fi
+fi


bin/before_install
```diff
@@ -18,3 +18,20 @@ elif [ ! -d "$spec_manageiq" ]; then
...
+  if [[ "$GITHUB_REPOSITORY_OWNER" = "ManageIQ" && "$GITHUB_BASE_REF" != "master" && "$GITHUB_REF_NAME" != "master" && "$GITHUB_REF_NAME" != "dependabot/"* ]]; then
+    echo "== Setup Gemfile.lock.release =="
+    cp -f "$gem_root/Gemfile.lock.release" "$gem_root/Gemfile.lock"
+
+    echo "== Patching Gemfile.lock... =="
+    `$gem_root/bin/patch_gemfile_lock`
+    echo "== Patching Gemfile.lock...Complete $? =="
+    echo
+  fi
+fi
```
jrafanie, line 37: Is the thought to move this into core logic with a flag to "patch with Gemfile.lock.release?"  It seems like most plugins would want to do this, right?

agrare: Correct the script to patch the Gemfile.lock will live in core but due to the cross-repo issues I have it here for now.
I hadn't thought about moving the entire `if [[ "$GITHUB_REPOSITORY_OWNER" = "ManageIQ" && "$GITHUB_BASE_REF" != "master" etc...` to a script in core we could definitely look into that since we have core cloned already at this point

bin/before_install
```diff
@@ -18,3 +18,18 @@ elif [ ! -d "$spec_manageiq" ]; then
...
+  echo "== Repository Owner: $GITHUB_REPOSITORY_OWNER =="
+  echo "== Base Ref: $GITHUB_BASE_REF =="
+  echo "== Ref Name: $GITHUB_REF_NAME =="
+
+  # Gemfile.lock.release only applies to non-master branches and PRs to non-master branches
+  if [[ "$GITHUB_REPOSITORY_OWNER" = "ManageIQ" && "$GITHUB_BASE_REF" != "master" && "$GITHUB_REF_NAME" != "master" && "$GITHUB_REF_NAME" != "dependabot/"* ]]; then
+    echo "== Setup Gemfile.lock.release =="
+    cp -f "$spec_manageiq/Gemfile.lock.release" "$gem_root/Gemfile.lock"
+
+    `$spec_manageiq/bin/ci/patch_plugin_gemfile_lock`
```
Fryguy, line 32: Why is this wrapped in backticks?  Since this is just a script, we should be able to just call it directly.

bin/before_install
```diff
@@ -18,3 +18,18 @@ elif [ ! -d "$spec_manageiq" ]; then
   echo "== Cloning manageiq sample app =="
   git clone https://github.com/ManageIQ/manageiq.git --branch oparin --depth 1 "$spec_manageiq"
 fi
+
+if [ -n "$CI" ]; then
+  echo "== Repository Owner: $GITHUB_REPOSITORY_OWNER =="
+  echo "== Base Ref: $GITHUB_BASE_REF =="
+  echo "== Ref Name: $GITHUB_REF_NAME =="
```
Fryguy, line 25: I think we should remove these before merge.

agrare: Forgot to mark this as WIP, this is only for testing the script since it will live in core but we can't cross-repo this yet

PR #171, opened 2022-08-29
append event.fixed value to event.event_type

OrGur1987:
since there's no `fixed` field in `event_streams` model, we prepend the `fixed` value (passed from autosde) to `event_type` field

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
@@ -31,9 +31,13 @@ class ManageIQ::Providers::Autosde::StorageManager::EventCatcher::Runner < Manag
 
   private
 
+  def parse_event_type(event)
+    event.fixed == 'yes' ? "#{event.event_type}_fixed" : event.event_type
+  end
+
   def event_to_hash(event, ems_id)
     {
-      :event_type               => event.event_type,
+      :event_type               => parse_event_type(event),
       :source                   => "AUTOSDE",
       :ems_ref                  => event.event_id,
       :physical_storage_ems_ref => event.storage_system,


app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
```diff
@@ -31,9 +31,14 @@ def queue_event(event)
 
   private
 
+  def prepend_fixed_to_event_type(event)
+    is_fixed = event.fixed == 'yes' ? 'fixed' : 'not fixed'
+    "#{is_fixed}: #{event.event_type}"
```
agrare, line 36: What are the event types?  Does it make sense to have `not fixed` in front of all of them?
E.g. I see `AUTOSDE_instance_power_on` I don't think that `not fixed: AUTOSDE_instance_power_on` makes sense.

How about
```suggestion
    event.fixed == 'yes' ? "#{event.event_type}_fixed" : event.event_type
```

app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
```diff
@@ -31,9 +31,14 @@ def queue_event(event)
 
   private
 
+  def prepend_fixed_to_event_type(event)
```
agrare, line 34: I think something like `parse_event_type` would be a better method name

app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
OrGur1987: I agree, the thing is appending `fixed` makes it invisible in the chart legend.

`fixed` appended:
image

'fixed' prepended (ignore `unfixed`, I agree it's redundant):
image
 
I know this problem belongs to a different repo/PR: https://github.com/ManageIQ/manageiq/pull/22081, 
but that's the whole reason for this change..
If I could control react carbon graph options through the yaml I would try to sort out the legend, but it's unclear how to do so (as I also wrote on Gitter).

agrare (approved):

PR #172, opened 2022-08-30
wait to refresh until status of job is success

Lauferin:
Until some time ago, the deletion and update of volumes, hosts, storages (and more) were done sinchronously, and when the task finished a refresh was done.

Since the update and deletion of those started being done with Celery, the refresh is being done immediately after the enqueue of the task without waiting for the task to finish to refresh.

I changed this to wait until the task successes and only then do the refresh (and in case it doesn't succeed or that the time waiting is longer than X seconds, it doesn't refresh but instead it raises an exception.

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
@@ -96,4 +96,24 @@ class ManageIQ::Providers::Autosde::StorageManager::AutosdeClient < AutosdeOpena
       # when set to nil, url refers to config.host, as it should be
     end
   end
+
+  private_class_method def self.wait_for_success(ems, task_id, time_to_sleep = 1, wait_time = 60)
+    task_status = nil
+    start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    while task_status != "SUCCESS" and task_status != "FAILURE" and Process.clock_gettime(Process::CLOCK_MONOTONIC) < start_time + wait_time
+      sleep(time_to_sleep)
+      task_status = ems.autosde_client.JobApi.jobs_pk_get(task_id).status
+    end
+    task_status
+  end
+
+  private_class_method def self.raise_non_success_exception(status)
+    if status == "FAILURE"
+      raise _("The Job failed")
+    elsif status == "PENDING" or status == "STARTED"
+      raise _("Timeout")
+    else
+      raise _("An unknown exception has occurred (Job status: '%{status_name}'") % {:status_name => status}
+    end
+  end
 end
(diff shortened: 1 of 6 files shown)


agrare:
@Lauferin we actually have a state machine pre-built which is intended to do exactly this, kick off some task, wait for it to finish, then kick of a refresh and wait for that to complete.

https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/ems_refresh_workflow.rb

This is preferable because your current method will potentially lock up an entire worker for up to a minute where a state machine will do a quick check then queue a state and return the worker to be able to perform other tasks.

app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
```diff
@@ -96,4 +96,15 @@ def configure_openapi_client
...
+  private_class_method def self.enqueue_refresh(target_class, target_id, ems_id, task_id)
+    ManageIQ::Providers::Autosde::StorageManager::EmsRefreshWorkflow.create_job(
+      :target_class   => target_class,
+      :target_id      => target_id,
+      :ems_id         => ems_id,
+      :native_task_id => task_id
+    ).tap do |job|
+      job.signal(:start)
+    end
+  end
```
agrare, line 109: I don't think it makes sense to have this in the `AutosdeClient` class since it isn't related to calling the API.  If you just want a "nicer" looking method you could always define a method on the `ManageIQ::Providers::Autosde::StorageManager::EmsRefreshWorkflow` class that takes positional args and calls start automatically.

app/models/manageiq/providers/autosde/storage_manager/ems_refresh_workflow.rb
```diff
@@ -0,0 +1,47 @@
+class ManageIQ::Providers::Autosde::StorageManager::EmsRefreshWorkflow < ManageIQ::Providers::EmsRefreshWorkflow
+
```
agrare, line 2: re: https://github.com/ManageIQ/manageiq-providers-autosde/pull/172/files#r963691043

```suggestion
  class << self
    def start_refresh(target_class, target_id, ems_id, task_id, options = {})
      options.reverse_merge!(:target_class => target_class, :target_id => target_id, :ems_id => ems_id, :native_task_id => task_id)
      ManageIQ::Providers::Autosde::StorageManager::EmsRefreshWorkflow.create_job(options).tap { |job| job.signal(:start) }
    end
  end
```

This should allow you to call:
```
ext_management_system.class::EmsRefreshWorkflow.start_refresh(self.class.name, id, ext_management_system.id, task_id)
```

agrare (approved):
I have a suggestion for some refactoring but nothing that should hold up this PR, LGTM

Lauferin:
Great. Soon I will make another PR including changes for create operations, and then we can make more refactoring.

PR #177, opened 2022-09-07
add create to queue

Lauferin:
When creating a volume or a storage system, we used to receive the object with its uuid and with it we refreshed. Then when we started with celery we were receiving an async object with a task_id, and we were not receiving the data of the object itself (which hasn't been created yet) and we had no way to find out about it.

Now the new feature in autosde saves the object_id reference in the task itself.

So now we can use the state machine of refresh to wait until the task is in SUCCESS state and then we do the refresh, like with delete and update.

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
@@ -16,23 +16,32 @@ class ManageIQ::Providers::Autosde::StorageManager::CloudVolume < ::CloudVolume
       :size    => options["size"],
       :count   => options["count"]
     )
+    task_id = ext_management_system.autosde_client.VolumeApi.volumes_post(vol_to_create).task_id
 
-    new_volume = ext_management_system.autosde_client.VolumeApi.volumes_post(vol_to_create)
-
-    EmsRefresh.queue_refresh(
-      InventoryRefresh::Target.new(
-        :manager     => ext_management_system,
-        :association => :cloud_volumes,
-        :manager_ref => {:ems_ref => new_volume.uuid}
-      )
-    )
+    options = {
+      :target_class   => :cloud_volumes,
+      :target_id      => nil,
+      :ems_id         => ext_management_system.id,
+      :native_task_id => task_id,
+      :interval       => 10.seconds,
+      :target_option  => "new"
+    }
+    ext_management_system.class::EmsRefreshWorkflow.create_job(options).tap { |job| job.signal(:start) }
   end
 
   # ================= delete  ================
 
   def raw_delete_volume
     task_id = ext_management_system.autosde_client.VolumeApi.volumes_pk_delete(ems_ref).task_id
-    ext_management_system.class::AutosdeClient.enqueue_refresh(self.class.name, id, ext_management_system.id, task_id)
+    options = {
+      :target_class   => self.class.name,
+      :target_id      => id,
+      :ems_id         => ext_management_system.id,
+      :native_task_id => task_id,
+      :interval       => 1.minute,
+      :target_option  => "existing"
+    }
+    ext_management_system.class::EmsRefreshWorkflow.create_job(options).tap { |job| job.signal(:start) }
   end
 
   # ================= edit  ================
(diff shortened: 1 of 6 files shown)


Lauferin:
@agrare , I've implemented the last change you wrote in PR #172 

This code has some refactoring to do, I'll write some of the things that I'm thinking about:

1) BEFORE you kill me, yes, I realize we have two calls to the client (one in poll_native_task and one in refresh) and the second one could be eliminated. And that's why there are two lines commented, in which I tried to add to the objects has the id info. But for some reason in the refresh method it doesn't identify the change, and I don't understand why.

2) Maybe one extra parameter which receives what we are updating (the ems, a new one or an existing one) could be more readable that writing that data in the target_id parameter.

3) It would be nice to pass as well the interval time between tasks in the queue (specially taking into account that the deletion of a volumes takes way more time that updating it)

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
```diff
@@ -17,22 +17,15 @@ def self.raw_create_volume(ext_management_system, options = {})
...
-
-    EmsRefresh.queue_refresh(
-      InventoryRefresh::Target.new(
-        :manager     => ext_management_system,
-        :association => :cloud_volumes,
-        :manager_ref => {:ems_ref => new_volume.uuid}
-      )
-    )
+    task_id = ext_management_system.autosde_client.VolumeApi.volumes_post(vol_to_create).task_id
+    ext_management_system.class::EmsRefreshWorkflow.start_refresh(:cloud_volumes, "new", ext_management_system.id, task_id)
```
agrare, line 21: I think we'd be better off passing `nil` rather than `"new"` here because there is code which will do a database query using `target_class+target_id` which would end up doing `CloudVolume.find("new")`

app/models/manageiq/providers/autosde/storage_manager/ems_refresh_workflow.rb
```diff
@@ -1,15 +1,23 @@
...
   alias start run_native_op
 
   def poll_native_task
-    case autosde_client.JobApi.jobs_pk_get(options[:native_task_id]).status
+    native_object = autosde_client.JobApi.jobs_pk_get(options[:native_task_id])
+    case native_object.status
     when "FAILURE"
       signal(:abort, "Task failed")
     when "SUCCESS"
+      # options[:native_object_id] = native_object.object_id
```
agrare, line 20: > But for some reason in the refresh method it doesn't identify the change, and I don't understand why.

If this is the line you are talking about the issue is you have to save the changes to the database, right here they are only modified in memory.

```suggestion
      options[:native_object_id] = native_object.object_id
      save!
```

Lauferin: Yes! that's exactly what I was talking about, and now you made me understand that for every refresh it looks for the job in the database and that it's not kept in memory.

So I squashed and tried it, it works well.

Do you suggest any other change? (or about the time interval...)

agrare:
> Maybe one extra parameter which receives what we are updating (the ems, a new one or an existing one) could be more readable thatnwriting that data in the target_id parameter.

Yes I think this is a good idea, we do have that extra `:options => {}` parameter that we could use instead of passing `"ems"` in the `target_id`

> It would be nice to pass as well the interval time between tasks in the queue (specially taking into account that the deletion of a volumes takes way more time that updating it)

You mean instead of always doing 1.minute?  Yes this is a good idea to reduce worker load when we don't need it checked as frequently.

Lauferin:
> Yes I think this is a good idea, we do have that extra `:options => {}` parameter that we could use instead of passing `"ems"` in the `target_id`

> You mean instead of always doing 1.minute? Yes this is a good idea to reduce worker load when we don't need it checked as frequently.

Hi, i did the changes. Since the line got too long I split it to two and now codeclimate complains about similar block codes. It also complains about it having more than 5 parameters (6). I can add any other of them also into options, but then it would look weird that some parameters are passed alone and others are passed into options. I thought about putting them all into options but in ticket #172 you preferred not to do that (but then we had fewer options). How would you prefer?

agrare:
@Lauferin yeah you can throw it all in options that is how it is being passed in to `Job` anyway I thought you wanted positional arguments based on your code in the last PR :laughing:

agrare (approved):

PR #178, opened 2022-09-12
Override where clause functions and added helpers dir

Oferlis:
In order to filter events by the selected storage system, overriding the where_clause functions was needed.
Since both models share most of the query structuring code, I added a new 'helpers' directory and a helper function that both models use. 

This PR is related to another in UI repo:
https://github.com/ManageIQ/manageiq-ui-classic/pull/8424

Other helpers should be added to this directory

### DIFF
--- a/app/helpers/manipulation_helper.rb
+++ b/app/helpers/manipulation_helper.rb
@@ -1,14 +1,14 @@
 module ManipulationHelper
-  def manipulate_storage_systems(assoc, storage_systems=nil)
-    if storage_systems.instance_of? Array and storage_systems.length >=1
+  def manipulate_storage_systems(assoc, storage_systems = nil)
+    if storage_systems.instance_of?(Array) && storage_systems.length >= 1
       rec_cond = ""
       storage_systems.each do |i|
-        rec_cond = rec_cond + "#{events_table_name(assoc)}.physical_storage_id = #{i}"
-        unless i.equal?storage_systems.last
-          rec_cond = rec_cond + " or "
+        rec_cond << "#{events_table_name(assoc)}.physical_storage_id = #{i}"
+        unless i.equal?(storage_systems.last)
+          rec_cond << " or "
         end
       end
-      return rec_cond
+      rec_cond
     end
   end
 end
--- a/app/helpers/manipulation_helper.rb
+++ b/app/helpers/manipulation_helper.rb
@@ -1,0 +1,14 @@
+module ManipulationHelper
+  def manipulate_storage_systems(assoc, storage_systems=nil)
+    if storage_systems.instance_of? Array and storage_systems.length >=1
+      rec_cond = ""
+      storage_systems.each do |i|
+        rec_cond = rec_cond + "#{events_table_name(assoc)}.physical_storage_id = #{i}"
+        unless i.equal?storage_systems.last
+          rec_cond = rec_cond + " or "
+        end
+      end
+      return rec_cond
+    end
+  end
+end
--- a/app/helpers/manipulation_helper.rb
+++ b/app/helpers/manipulation_helper.rb
@@ -1,14 +1,9 @@
 module ManipulationHelper
-  def manipulate_storage_systems(assoc, storage_systems = nil)
+  def storage_system_where_clause(assoc, storage_systems = nil)
     if storage_systems.instance_of?(Array) && storage_systems.length >= 1
-      rec_cond = ""
-      storage_systems.each do |i|
-        rec_cond << "#{events_table_name(assoc)}.physical_storage_id = #{i}"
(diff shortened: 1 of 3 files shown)


Fryguy:
Not sure why the bot is not triggering here, but the use of `and` should be `&&` instead.

app/helpers/manipulation_helper.rb
```diff
@@ -0,0 +1,14 @@
+module ManipulationHelper
```
Fryguy, line 1: I'm not sure this should be a top-level helper, because then it's global to the whole app...it should probably namespaced to the plugin.

app/helpers/manipulation_helper.rb
```diff
@@ -0,0 +1,14 @@
...
+  def manipulate_storage_systems(assoc, storage_systems=nil)
+    if storage_systems.instance_of? Array and storage_systems.length >=1
+      rec_cond = ""
+      storage_systems.each do |i|
+        rec_cond = rec_cond + "#{events_table_name(assoc)}.physical_storage_id = #{i}"
```
Fryguy, line 6: Prefer << over + (same below)

Fryguy:
Please address the rubocop issues found by the bot.

app/helpers/manipulation_helper.rb
```diff
@@ -0,0 +1,14 @@
...
+  def manipulate_storage_systems(assoc, storage_systems = nil)
+    if storage_systems.instance_of?(Array) && storage_systems.length >= 1
+      rec_cond = ""
+      storage_systems.each do |i|
+        rec_cond << "#{events_table_name(assoc)}.physical_storage_id = #{i}"
+        unless i.equal?(storage_systems.last)
+          rec_cond << " or "
+        end
+      end
+      rec_cond
```
Fryguy, line 11: Simpler:

```suggestion
      storage_systems.map do |i|
        "#{events_table_name(assoc)}.physical_storage_id = #{i}"
      end.join(" OR ")
```

app/helpers/manipulation_helper.rb
```diff
@@ -0,0 +1,14 @@
+module ManipulationHelper
+  def manipulate_storage_systems(assoc, storage_systems = nil)
+    if storage_systems.instance_of?(Array) && storage_systems.length >= 1
```
Fryguy, line 3: What is the storage_systems.instance_of?(Array) check actually checking for?  Are you really just checking for nil?

app/helpers/manipulation_helper.rb
```diff
@@ -0,0 +1,14 @@
+module ManipulationHelper
+  def manipulate_storage_systems(assoc, storage_systems = nil)
```
Fryguy, line 2: Naming wise, I'm not sure the term manipulate is clear here.  This seems to just generate a where clause, so perhaps

```suggestion
  def storage_systems_where_clause(assoc, storage_systems = nil)
```

app/models/manageiq/providers/autosde/storage_manager.rb
```diff
@@ -114,6 +114,14 @@ def count_resources(component = nil)
     end
   end
 
+  def event_where_clause(assoc = :ems_events, storage_systems = nil)
+    if storage_systems && storage_systems != ['']
```
Fryguy, line 118: What would cause `['']` to be an incoming value?  If you can prevent this from the calling side and accept an empty array instead, then this line can be simplified to the following since `.present?` handle both `nil` and `[]`

```ruby
   if storage_systems.present?
```

app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
```diff
@@ -13,6 +15,19 @@ def raw_delete_physical_storage
...
+  def event_where_clause(assoc = :ems_events, storage_systems = nil)
+    case assoc.to_sym
+    when :ems_events, :event_streams
+      if storage_systems && storage_systems != ['']
+        return manipulate_storage_systems(assoc, storage_systems)
+      end
+
+      ["#{events_table_name(assoc)}.physical_storage_id = ?", id]
+    when :policy_events
+      ["target_id = ? and target_class = ? ", id, self.class.base_class.name]
```
Fryguy, line 27: Super-minor, but capitalize and SQL keywords for consistency.

```suggestion
      ["target_id = ? AND target_class = ? ", id, self.class.base_class.name]
```

app/helpers/manipulation_helper.rb
Fryguy: Right, but there are other ways of doing that like checking `respond_to?(:length)`.  But what are you seeing coming in aside from the nil?  or is it only the nil?

PR #179, opened 2022-09-13
add provider events to Notifications

Lauferin:
We're receiving events from Autosde, which were only presented in the monitoring section.

Now, when polling the events, they're also being sent to Notifications, with a new notification type with the name of the type of the event.

Screen Shot 2022-09-13 at 13 22 31

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
@@ -25,8 +25,19 @@ class ManageIQ::Providers::Autosde::StorageManager::EventCatcher::Runner < Manag
 
   def queue_event(event)
     _log.info("#{log_prefix} Caught event AutoSDE [#{event.event_id}]")
+    add_notification_type(event)
     event_hash = event_to_hash(event, @cfg[:ems_id])
     EmsEvent.add_queue('add', @cfg[:ems_id], event_hash)
+  end
+
+  def add_notification_type(event)
+    NotificationType.create_with(
+      :name       => event.event_type,
+      :level      => "warning",
+      :audience   => "global",
+      :expires_in => 7.days,
+      :message    => "%{event_type} in event %{uuid}, error code: %{error_code}. %{description}"
+    ).find_or_create_by(:name => event.event_type)
   end
 
   private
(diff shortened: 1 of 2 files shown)


Lauferin:
Hi @agrare 

I added the notification type. A new one will be made for every new type that comes from the provider.

You told me to use emit_for_event instead of the rest of the code. I've temporally left the old code for adding the notification and wrote the new commented. The reason why I'm not being able to do it with the method is because emit_for_event demands receiving an event that has "full_data" and "target" properties, but the event object I'm getting from the provider doesn't have those (and I can't add them in the client, it's generated code and that would be a real mess).
If we were in Python I would add them dynamically but from what I've seen Ruby doesn't support that option (it supports adding it with instance_variable_set() but then in the function we would need to get it with instance_variable_get() ). I could make a new event class with getters and setters but I'm not sure if it's worth doing that only for using the emit_for_event function (if you prefer this way let me know).

agrare:
hey @Lauferin you are building the event hash in the EventCatcher `event_to_hash` method.  Typically `full_data` is the entire native event converted to a hash

Lauferin:
Hi @agrare yes, but emit_for_event is expecting an object with attributes, not a hash. Then I still need to add them as attributes to the event itself. So I ask if you agree that I should build there a whole new event object from scratch or maybe you're thinking of a different way?

agrare:
@Lauferin you don't have to call emit_for_event yourself, it is called automatically when the event is saved so it gets the actual event record

app/models/manageiq/providers/autosde/storage_manager/event_catcher/runner.rb
```diff
@@ -25,10 +25,21 @@ def monitor_events
...
+  def add_notification_type(event)
+    NotificationType.create_with(
+      :name       => event.event_type,
+      :level      => "warning",
+      :audience   => "global",
+      :expires_in => 7.days,
+      :message    => "%{event_type} in event %{uuid}, error code: %{error_code}. %{description}"
+    ).find_or_create_by(:name => event.event_type)
+  end
+
```
agrare, line 42: @Lauferin notification_types are seeded on startup so you don't have to create these on the fly.  They are seeded from https://github.com/ManageIQ/manageiq/blob/fd400ab58f82496f641ffa3f3147fc12e38345b7/db/fixtures/notification_types.yml

I would like to make this pluggable so you can define these in this repo but for now for testing you can add your event types you want notifications for to that yaml file.

Lauferin: I've already tested all what was needed... so I guess we freeze this PR until the pluggable is done?

agrare:
@Lauferin pluggable notification types are ready you can drop a yaml array in `content/notification_types.yml`

Lauferin:
Cool! pushed.

agrare (approved):

PR #185, opened 2022-10-26
change version to 2.0.0 after changes in the autosde api

eilam20:
We made changes in the autosde api.
Now we return a response with pagination, so we need to upgrade the gem version to 2.0.0


agrare:
@eilam20 curious how this passes with the new gem but without re-recording the VCRs if the API response is different?

eilam20:
This gem version will accept response from old autosde versions but we need to adjust it to the new paginated response that the new version of autosde will return.

agrare:
@eilam20 sounds good, do you want to split your VCRs into v1 and v2 to ensure both paths work?

eilam20:
@agrare hi I split the specs into two versions, can you check it?

agrare (approved):
LGTM!

PR #186, opened 2022-10-31
choose host initiator group on host creation

OrGur1987:
added a host initiator group field to host initiator form, which enables choosing an existing host group from the specified storage system, or choose None for no group.

this also requires a change in classic-ui repo:
https://github.com/ManageIQ/manageiq-ui-classic/pull/8500

the new field:

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
@@ -11,14 +11,18 @@ class ManageIQ::Providers::Autosde::StorageManager::HostInitiator < ::HostInitia
     iqns = Array(options['iqn'])
     iqn_values = iqns.join(":")
 
+    host_cluster_name = options.dig('host_initiator_group', 'label')
+    host_cluster_name = '' if host_cluster_name == '<None>'
+
     host_initiator_to_create = ext_management_system.autosde_client.StorageHostCreate(
-      :name           => options['name'],
-      :port_type      => options['port_type'],
-      :storage_system => PhysicalStorage.find(options['physical_storage_id']).ems_ref,
-      :iqn            => iqn_values || "",
-      :wwpn           => wwpn_values || "",
-      :chap_name      => options['chap_name'] || "",
-      :chap_secret    => options['chap_secret'] || ""
+      :name              => options['name'],
+      :port_type         => options['port_type'],
+      :storage_system    => PhysicalStorage.find(options['physical_storage_id']).ems_ref,
+      :host_cluster_name => host_cluster_name,
+      :iqn               => iqn_values || "",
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
```diff
@@ -11,14 +11,19 @@ def self.raw_create_host_initiator(ext_management_system, options = {})
     iqns = Array(options['iqn'])
     iqn_values = iqns.join(":")
 
+    def self.parse_cluster_name(cluster_name)
+      cluster_name == '<None>' ? '' : cluster_name
+    end
+
```
agrare, line 17: Need to move this out of the `raw_create_host_initiator` method definition.  Also it probably should be a `private_class_method`

OrGur1987: the code climate check suggested using a lambda, hope that's ok

agrare: Honestly I think a simple local variable would be easier

app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
```diff
@@ -11,14 +11,17 @@ def self.raw_create_host_initiator(ext_management_system, options = {})
     iqns = Array(options['iqn'])
     iqn_values = iqns.join(":")
 
+    parse_cluster_name = -> (cluster_name) { cluster_name == '<None>' ? '' : cluster_name }
+
```
agrare, line 15:
```suggestion
    cluster_name = options.dig('host_initiator_group', 'label')
    cluster_name = '' if cluster_name == '<None>'
```

app/models/manageiq/providers/autosde/storage_manager/host_initiator.rb
```diff
@@ -11,14 +11,17 @@ def self.raw_create_host_initiator(ext_management_system, options = {})
...
-      :port_type      => options['port_type'],
-      :storage_system => PhysicalStorage.find(options['physical_storage_id']).ems_ref,
-      :iqn            => iqn_values || "",
-      :wwpn           => wwpn_values || "",
-      :chap_name      => options['chap_name'] || "",
-      :chap_secret    => options['chap_secret'] || ""
+      :name              => options['name'],
+      :port_type         => options['port_type'],
+      :storage_system    => PhysicalStorage.find(options['physical_storage_id']).ems_ref,
+      :host_cluster_name => parse_cluster_name.(options['host_initiator_group']['label']),
```
agrare, line 20:
```suggestion
      :host_cluster_name => host_cluster_name
```

agrare (approved):

PR #188, opened 2022-11-09
refresh name from volume_name

Lauferin:
There has been a change in the volume model of autosde. The field 'name' in volume now includes the storage system and pool name, and the specific name is now in field volume_name, which is the one that manageIQ should show. Not doing that leaves two problems: 1) it doesn't look good (see image below) and 2) autosde's rest receives a name and adds to it the pool and storage, so if we send it already with that data the name just grows on and on...

Screen Shot 2022-11-02 at 14 10 47


spec/models/manageiq/providers/autosde/storage_manager/refresher_spec.rb
```diff
@@ -135,7 +135,7 @@
 
         assert_inventory_not_changed { run_targeted_refresh(ems.cloud_volumes.find_by(:ems_ref => "ac287c2d-1776-48a3-a5c9-06327f4a57c4")) }
       end
-
+      z
```
agrare, line 138:
```suggestion
```

Lauferin: Sorry, I was trying to resolve the tests' problems and apparently because of the ctrl+z...

Anyway, I'm having a hard time solving them (specially when I see that everything works fine). I understand that the data it's not really doing an API call so I'm trying to solve this from the cassettes, although I don't see anywhere inside them "new-volume".

agrare: @Lauferin it looks like the issue is just updating the tests to expect a name on the cloud volume rather than `nil` as it does currently

Lauferin:
Ok, now tests work :-)

PR #189, opened 2022-11-14
return error when there is no 'response_body'

Lauferin:
When trying to add a storage from a wrong IP, we get the following response to the UI:

9ad9dad8-178b-4316-9bc5-2a9d011af179

Now it will receive the original error of the response of the HTTP request:

Screen Shot 2022-11-14 at 12 51 55

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
@@ -70,8 +70,12 @@ class ManageIQ::Providers::Autosde::StorageManager::AutosdeClient < AutosdeOpena
     data = self.AuthenticationApi.token_auth_post(auth_request, opts)
     @token = data.token
   rescue AutosdeOpenapiClient::ApiError => e
-    error_data = JSON.parse(e.response_body)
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/autosde/storage_manager/autosde_client.rb
```diff
@@ -70,8 +70,11 @@ def login
     data = self.AuthenticationApi.token_auth_post(auth_request, opts)
     @token = data.token
   rescue AutosdeOpenapiClient::ApiError => e
-    error_data = JSON.parse(e.response_body)
-    raise AUTH_ERRR_MSG + error_data.dig("detail", "non_field_errors")&.first
+    error_data = e.to_s
+    if e.response_body
+      error_data = JSON.parse(e.response_body).dig("detail", "non_field_errors")&.first
+    end
```
agrare, line 76: Hm if you're going to overwrite `error_data` maybe better to do something like:
```suggestion

    error_data = if e.response_body
      JSON.parse(e.response_body).dig("detail", "non_field_errors")&.first
    else
      e.to_s
    end
```

agrare (approved):

Lauferin:
PR:
https://github.com/ManageIQ/manageiq-providers-autosde/pull/189

PR #192, opened 2022-11-30
check for autosde refresh status after creating storage system

Lauferin:
After adding a storage system, the refresh workflow takes care of refreshing after autosde's task of adding the storage system gets to success state. That's also the case for creating / updating / deleting volumes / hosts / mappings

But the problem is that, after adding a storage system, autosde shoots another asynchronous task of refreshing itself (which brings resources from the actual storage). And we want to refresh miq only after that task is completed, so it also refreshes pools, volumes and hosts, and so it doesn't look like this:

cf42aa6f-1d1b-4ba5-99d2-68196358b09c

So I added some data to the refresh tasks to be able to locate the task, and updated the refresh workflow with a second poll task for this specific case.

I could have done it in the same poll_native_task, but in that case I would have to asked conditions at the beginning of it for every refresh task called, which seems to me wasteful. So I did it in a second function (and that's why I overrode the load_transtions, in order to add the transition for that new function).

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/ems_refresh_workflow.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/ems_refresh_workflow.rb
@@ -13,13 +13,44 @@ class ManageIQ::Providers::Autosde::StorageManager::EmsRefreshWorkflow < ManageI
     when "SUCCESS"
       options[:native_object_id] = native_object.object_id
       save!
-      queue_signal(:refresh)
+      if options[:target_option] == 'new' && options[:target_class] == :physical_storages
+        options[:target_option] = "ems"
+        save!
+        queue_signal(:poll_refresh_task, :deliver_on => Time.now.utc + options[:interval])
+      else
+        queue_signal(:refresh)
+      end
     else
       queue_signal(:poll_native_task, :deliver_on => Time.now.utc + options[:interval])
     end
   rescue => err
     _log.log_backtrace(err)
     signal(:abort, err.message, "error")
+  end
+
+  def poll_refresh_task
+    opts = {
+      :query_params => {
+        :object_id => options[:native_object_id]
+      }
+    }
+    tasks = autosde_client.JobApi.jobs_get(opts)
+    raise _('No refresh task found') if tasks.count < 1
+
+    task = tasks.detect { |t| t.task_name.match?("update_system") }
+    raise _('No refresh task found') if task.nil?
+
+    case task.status
+    when "FAILURE"
+      signal(:abort, "Refresh failed")
+    when "SUCCESS"
+      queue_signal(:refresh)
+    else
+      queue_signal(:poll_refresh_task, :deliver_on => Time.now.utc + options[:interval])
+    end
+  rescue => err
+    _log.log_backtrace(err)
+    signal(:abort, err.message, opts)
   end
 
   def refresh
@@ -47,6 +78,24 @@ class ManageIQ::Providers::Autosde::StorageManager::EmsRefreshWorkflow < ManageI
 
       queue_signal(:poll_refresh)
     end
+  end
+
+  def load_transitions
+    self.state ||= 'initialize'
+
+    {
+      :initializing      => {'initialize'       => 'waiting_to_start'},
+      :start             => {'waiting_to_start' => 'running'},
+      :poll_native_task  => {'running'          => 'running'},
+      :poll_refresh_task => {'running'          => 'running'},
+      :refresh           => {'running'          => 'refreshing'},
+      :poll_refresh      => {'refreshing'       => 'refreshing'},
+      :post_refresh      => {'refreshing'       => 'post_refreshing'},
+      :finish            => {'*'                => 'finished'},
+      :abort_job         => {'*'                => 'aborting'},
+      :cancel            => {'*'                => 'canceling'},
+      :error             => {'*'                => '*'}
+    }
   end
 
   def autosde_client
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/autosde/storage_manager/ems_refresh_workflow.rb
```diff
@@ -22,6 +28,26 @@ def poll_native_task
...
   end
 
+  def poll_refresh_task
+    opts = {
+      :query_params => {
+        :object_id => options[:native_object_id]
+      }
+    }
+    tasks = autosde_client.JobApi.jobs_get(opts)
+    case tasks.first.status
```
agrare, line 38: I think you need to handle if tasks is `[]` also otherwise this will raise an exception.

agrare: Is there a job-type or naming convention that we can use to identify the "refresh task" versus some other task?  What happens if another task for that object_id kicks off that finishes quickly before the self-refresh task is done?

agrare (commented):
Overall looks good just the two comments

app/models/manageiq/providers/autosde/storage_manager/ems_refresh_workflow.rb
Lauferin: Good question. There is, and I actually wrote it at the beginning, but then I decided to remove it since it makes a tide coupling between autosde and miq. It's actually a name that could change very easily and break everything, so I didn't like the idea.

And then comes your question, what happens if another task finishes before that refresh is done. The answer is that it shouldn't happen, because we're speaking here about the creation of the storage, and the user can't really see anything until that first refresh task is complete. And even in an extreme scenario where that does happen for some reason, what will happen is that the refresh will wait for that new task, and not only I don't mind that but actually it's preferable, so that the refresh also watches after that change as well.

Lauferin: > I think you need to handle if tasks is [] also otherwise this will raise an exception.

I don't understand, the whole function is handled, it will get to the abort... or do you mean to handle it individually and give it a more friendly error message?

Lauferin: > > I don't understand, the whole function is handled, it will get to the abort... or do you mean to handle it individually and give it a more friendly error message?
> 
> I'm thinking if somehow we check before the task is created, this will blow up and abort as opposed to retry.
> 
> Plus if you want to require there to be exactly 1 result in an array it is better to check that and raise a failure with a proper error message rather than a confusing `NoMethodError first for NilClass`

Now I got it. But to be honest, I actually prefer it to abort rather than to retry. Because synchronously speaking, the autosde's refresh task is being fired before the create task is finished (and then it behaves asynchronously, but it the refresh task isn't saved, the autosde's refresh won't success).
So technically, this can't happen. And if for some strange reason it does happen, if the task wasn't created till then, it won't be created at all, so there is no reason to retry. I do agree with the proper error message. I don't think we need it because I don't see a scenario of that happening, but if you want I'll add it.

Lauferin: Done. But I wrote < 1, because it's supposed to be 2 (it will also find the task of the creation of the storage system, which has the same object_id)

By the way, I changed the time of the delete of storage, it takes way less time that one minute, if it's Ok :)

agrare: Okay highly discourage using `.first` on an array with 2 items without first sorting or selecting on some unique value.  Is there some attribute about the storage system creation task that we can use to exclude it or something about the refresh task we can select on?

`tasks.reject { |t| t.name.match?("Create Storage System") }.first`
or
`tasks.detect { |t| t.name.match("Refresh task") }`

Lauferin: That's what I wanted to avoid in the first place, but now that you say, If we bring everything and then filter it here, it makes more sense. So we're looking for "update_system", that's not supposed to change.
So cool.
Do you want me to add:
`raise _('No refresh task found') if task == nil`
?

agrare: Just checking, could you use `"update_system"` in your query_params?

Lauferin: I can use "autosde_site_manager.engine.tasks.update_system" but that's what I told you that I wanted to avoid because of the coupling. And I definitely don't want to allow the REST to find something by part of its name by standard...

app/models/manageiq/providers/autosde/storage_manager/ems_refresh_workflow.rb
```diff
@@ -22,6 +28,29 @@ def poll_native_task
...
+    opts = {
+      :query_params => {
+        :object_id => options[:native_object_id]
+      }
+    }
+    tasks = autosde_client.JobApi.jobs_get(opts)
+    raise _('No refresh task found') if tasks.count < 1
+    task = tasks.detect { |t| t.task_name.match("update_system") }
+    raise _('No refresh task found') if task.nil?
+    case task.status
```
agrare, line 41: Two minor things, 1. you want `match?` not `match` because all you care about is the boolean return and 2. rubocop will warn about not having a newline after the raise
```suggestion
    tasks = autosde_client.JobApi.jobs_get(opts)
    raise _('No refresh task found') if tasks.count < 1

    task = tasks.detect { |t| t.task_name.match?("update_system") }
    raise _('No refresh task found') if task.nil?

    case task.status
```

Also just checking that the `task_name` isn't literally `"update_system"`?  If so you could do a `t.task_name == "update_system"`

agrare: Okay I see from above the name is `autosde_site_manager.engine.tasks.update_system` so good with `.match?`

agrare (approved):

PR #193, opened 2022-12-13
add ems_refresh_workflow to host group and updated times to wait

Lauferin:
The same way we're refreshing after all create / update / delete in volumes / hosts / physical storage, through the ems_refresh_workflow (which refreshes only after their tasks gets to success state), we're doing now for host initiator groups.

Screen Shot 2022-12-13 at 17 46 25

Besides, I reduced the amount of time to wait be
tween attempts of refreshing for other tasks, since we've reduced the time needed for those operations (and it will help us reduce time of testing with testcafe).

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/host_initiator_group.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/host_initiator_group.rb
@@ -6,7 +6,17 @@ class ManageIQ::Providers::Autosde::StorageManager::HostInitiatorGroup < ::HostI
       :name           => options['name'],
       :storage_system => PhysicalStorage.find(options['physical_storage_id']).ems_ref
     )
-    ext_management_system.autosde_client.HostClusterApi.host_clusters_post(host_initiator_group_to_create)
-    EmsRefresh.queue_refresh(ext_management_system)
+    task_id =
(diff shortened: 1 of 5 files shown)


app/models/manageiq/providers/autosde/storage_manager/host_initiator_group.rb
```diff
@@ -6,7 +6,17 @@ def self.raw_create_host_initiator_group(ext_management_system, options = {})
...
-    EmsRefresh.queue_refresh(ext_management_system)
+    task_id =
+      ext_management_system.autosde_client.HostClusterApi.host_clusters_post(host_initiator_group_to_create).task_id
+
+    options = {
+      :target_class   => nil,
+      :target_id      => nil,
+      :ems_id         => ext_management_system.id,
+      :native_task_id => task_id,
+      :interval       => 10.seconds,
```
agrare, line 17: Any reason this is 10sec where the rest are 20sec?

Lauferin: all the post are 10 and delete 20

agrare (approved):

agrare:
Not sure why the github actions run isn't showing up in the checks section here but it is green https://github.com/ManageIQ/manageiq-providers-autosde/actions/runs/3687001871

PR #194, opened 2023-01-01
storage manager volume create: basic and advanced modes

OrGur1987:
filter services and resources by capabilities on volume create:

Basic: filter services by required capabilities and choose an existing service from the filtered list.
Advanced: choose a name for a new service, filter resources by required capabilities, and choose resources to attach to the new service from the filtered list. the new volume(s) will be created for the new service, on one of the attached resources.

changes:
- cloud_volume.rb: 
  -  on params_for_create: get capabilities from provider and use them as options in a multiSelect field. example of capabilities json field from providers api:
image
  - on raw_create_volume: build the creation hash according to the selected mode.
- inventory:
  - collect storage manager capabilities from autosde and save them in the providers capabilities json field.
  - collect service and resource capabilities from autosde and save in the capabilities json field in their respective model in miq.

related PR in ui-classic:
https://github.com/ManageIQ/manageiq-ui-classic/pull/8582

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
@@ -10,12 +10,21 @@ class ManageIQ::Providers::Autosde::StorageManager::CloudVolume < ::CloudVolume
 
   def self.raw_create_volume(ext_management_system, options = {})
     # @type [StorageService]
-    vol_to_create = ext_management_system.autosde_client.VolumeCreate(
-      :service => ext_management_system.storage_services.find(options["storage_service_id"]).ems_ref,
+    creation_hash = {
+      :service => "",
       :name    => options["name"],
       :size    => options["size"],
       :count   => options["count"]
-    )
+    }
+
+    if options['mode'] == 'Basic'
+      creation_hash[:service] = ext_management_system.storage_services.find(options["storage_service_id"]).ems_ref
+    else
+      creation_hash[:service_name] = options["new_service_name"]
+      creation_hash[:resources] = ext_management_system.storage_resources.find(options["storage_resource_id"].to_a).pluck(:ems_ref)
+    end
+
+    vol_to_create = ext_management_system.autosde_client.VolumeCreate(creation_hash)
     task_id = ext_management_system.autosde_client.VolumeApi.volumes_post(vol_to_create).task_id
 
     options = {
@@ -100,21 +109,12 @@ class ManageIQ::Providers::Autosde::StorageManager::CloudVolume < ::CloudVolume
   end
 
   def self.params_for_create(provider)
-    services = provider.storage_services.map { |service| {:value => service.id.to_s, :label => service.name} }
+    capabilities = provider.capabilities.map do |capability|
+      {:label => "#{capability['abstract_capability']}: #{capability['value']}", :value => capability['uuid']}
+    end
 
     {
       :fields => [
-        {
-          :component    => "select",
-          :name         => "storage_service_id",
-          :id           => "storage_service_id",
-          :label        => _("Storage Pool"),
-          :isRequired   => true,
-          :validate     => [{:type => "required"}],
-          :options      => services,
-          :includeEmpty => true,
-          :isDisabled   => false
-        },
         {
           :component  => "text-field",
           :id         => "size",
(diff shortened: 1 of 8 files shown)


agrare:
Hey @OrGur1987 can you include an example of what the capabilities json looks like?

OrGur1987:
> Hey @OrGur1987 can you include an example of what the capabilities json looks like?

edited description

OrGur1987:
Hi @agrare tests are successful, it's ready for a review

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
```diff
@@ -10,12 +10,24 @@ class ManageIQ::Providers::Autosde::StorageManager::CloudVolume < ::CloudVolume
...
-      :service => ext_management_system.storage_services.find(options["storage_service_id"]).ems_ref,
+    creation_hash = {
+      :service => "",
       :name    => options["name"],
       :size    => options["size"],
       :count   => options["count"]
-    )
+    }
+
+    if options['mode'] == 'Basic'
```
agrare, line 20: Where is `options['mode']` coming from?  I don't see it in the DDF params

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
```diff
@@ -10,12 +10,24 @@ class ManageIQ::Providers::Autosde::StorageManager::CloudVolume < ::CloudVolume
...
+      creation_hash[:service] = ext_management_system.storage_services.find(options["storage_service_id"]).ems_ref
```
agrare, line 21: If `options["storage_service_id"]` is an array (which the lack of any `.to_a` down on [line 25](https://github.com/ManageIQ/manageiq-providers-autosde/pull/194/files#diff-54d13cf1217007ed4868559270a4b2861045c10b072508e2d5ae8d3e41f61f20R25) would indicate) then this `.find([id])` will return an array and `.ems_ref` will fail

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
```diff
@@ -10,12 +10,24 @@ class ManageIQ::Providers::Autosde::StorageManager::CloudVolume < ::CloudVolume
...
+    else
+      creation_hash[:service_name] = options["new_service_name"]
+      creation_hash[:resources] = []
+      options["storage_resource_id"].each do |resource|
```
agrare, line 25: Should we `Array.wrap()` or `.to_a` this just to be safe?

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
```diff
@@ -10,12 +10,24 @@ class ManageIQ::Providers::Autosde::StorageManager::CloudVolume < ::CloudVolume
...
+        creation_hash[:resources].push(ext_management_system.storage_resources.find(resource['value']).ems_ref)
+      end
```
agrare, line 27: A more efficient way of doing this would be
```suggestion
      creation_hash[:resources] = ext_management_system.storage_resources.find(options["storage_resource_id"].to_a).pluck(:ems_ref)
```

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
```diff
@@ -100,20 +112,22 @@ def params_for_update
   end
 
   def self.params_for_create(provider)
-    services = provider.storage_services.map { |service| {:value => service.id.to_s, :label => service.name} }
+    capabilities = []
+    provider.capabilities.map do |capability|
+      capabilities.push({:label => "#{capability['abstract_capability']}: #{capability['value']}", :value => capability['uuid']})
+    end
```
agrare, line 118: You already are using `.map` so might as well just assign the result
```suggestion
    capabilities = provider.capabilities.map do |capability|
      {:label => "#{capability['abstract_capability']}: #{capability['value']}", :value => capability['uuid']}
    end
```

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
OrGur1987: it used to be in `cloud-volume-form.schema.js` but I moved it to `cloud_voume.rb` following @Fryguy's request: 
https://github.com/ManageIQ/manageiq-ui-classic/pull/8582#issuecomment-1379580040

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
OrGur1987: options["storage_service_id"] is a single uuid string value, passed from the non-multi-select field `storage_service_id`.  

options["storage_resource_id"] is an array of uuid strings.

agrare (approved):

PR #195, opened 2023-01-03
Added capabilities field to storage resources object

galoiring:
Inside parser added capabilities field so it can be listed in the textual summary page. 

related pr:
 https://github.com/ManageIQ/manageiq-ui-classic/pull/8584
                  https://github.com/ManageIQ/manageiq-schema/pull/677


app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
```diff
@@ -51,7 +51,8 @@ def storage_resources
         :ems_ref          => resource.uuid,
         :logical_free     => resource.logical_free,
         :logical_total    => resource.logical_total,
-        :physical_storage => persister.physical_storages.lazy_find(resource.storage_system)
+        :physical_storage => persister.physical_storages.lazy_find(resource.storage_system),
+        :capabilities     => resource.capability_values_json
```
agrare, line 55: Since we have a jsonb column why not parse this into an easily consumable hash?

galoiring: parse it here with JSON.parse ?

galoiring: I think we need to confirm this with @ereztheodorou , 
this field might be use for other things.

galoiring: @agrare ok. 
so we will parse those column here..

app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
```diff
@@ -44,14 +44,22 @@ def physical_storages
     end
   end
 
+  def parse_capabilities(capabilities)
+    caps = JSON.parse(capabilities)
```
Fryguy, line 48: I'm not sure I understand this - I thought that using jsonb would _automatically_ get capabilities to be a deserialized Hash.

app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
```diff
@@ -46,12 +46,15 @@ def physical_storages
 
   def storage_resources
     collector.storage_resources.each do |resource|
+      # require 'byebug'
+      # byebug
```
agrare, line 50:
```suggestion
      # require 'byebug'
      # byebug
```

agrare (commented):
Looks good just drop the byebug comment

app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
galoiring: 🤦

agrare (approved):

PR #199, opened 2023-01-24
Added post refresh function to fill empty event columns

Oferlis:
Since the activation of the events worker was when the ems was set up, events were received without a physical storage name, which led to errors in the events chart in the EMS dashboard.
The fix was to add a condition to the start of the events worker, so the polling would start when physical storage is added to the EMS.

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager.rb
@@ -302,10 +302,10 @@ class ManageIQ::Providers::Autosde::StorageManager < ManageIQ::Providers::Storag
 
     events.each do |e|
       physical_storage_uuid = e.full_data[:storage_system]
-      physical_storage = PhysicalStorage.where(:ems_ref => physical_storage_uuid).to_a
+      physical_storage = PhysicalStorage.find_by(:ems_id => ems_id, :ems_ref => physical_storage_uuid)
 
-      e.physical_storage_name = physical_storage[0].name
-      e.physical_storage_id = physical_storage[0].id
+      e.physical_storage_name = physical_storage.name
+      e.physical_storage_id = physical_storage.id
       e.save
     end
   end
--- a/app/models/manageiq/providers/autosde/storage_manager.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager.rb
@@ -298,7 +298,7 @@ class ManageIQ::Providers::Autosde::StorageManager < ManageIQ::Providers::Storag
   end
 
   def self.post_refresh_ems(ems_id, _)
-    events = EmsEvent.where("ems_id = ? AND (physical_storage_id IS NULL OR physical_storage_name IS NULL) LIMIT NULL", ems_id)
+    events = EmsEvent.where("ems_id = ? AND (physical_storage_id IS NULL OR physical_storage_name IS NULL)", ems_id)
 
     events.each do |e|
       physical_storage_uuid = e.full_data[:storage_system]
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/autosde/storage_manager/event_catcher/stream.rb
```diff
@@ -17,6 +17,12 @@ def stop
   end
 
   def poll_events(&block)
+    physical_storages = []
+    while physical_storages.empty?
+      physical_storages = PhysicalStorage.where(:ems_id => @ems.id)
+      sleep(poll_sleep)
+    end
```
agrare, line 24: Hey @Oferlis so this has a number of issues with it that we can get into but just so I understand, is the problem you're trying to solve that events are caught by the event catcher and saved by the event handler before the refresh_worker finishes the first refresh leaving events without an associated `physical_storage_id`?  Usually the way this is solved is by adding a `post_create` action to the model which reconnects events which were saved before the associated record is created.  See `VmOrTemplate#reconnect_events` for an example of this https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_or_template.rb#L862

This does mean that there has to be something unique saved to the event_streams record that can be used, so I think we'd want to save the ems_ref then we can use that for reconnecting in the future.

Oferlis: Hey @agrare when should I call the post_create method? 
Should it be a part of the event polling loop?

agrare: @Oferlis it is part of refresh (because at that point you know the inventory records are created).

agrare: https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/cloud_manager/refresher.rb#L15-L18 is an example of telling refresh to post-process those classes
Then you have to define a `self.post_refresh_ems` method on the class like https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_or_template.rb#L1068

Oferlis:
Hey @agrare, this is the new approach that worked well. 
I will do further testing, but does this solution look better?

agrare:
@Oferlis :+1: yes much, good work

app/models/manageiq/providers/autosde/storage_manager.rb
```diff
@@ -296,4 +296,17 @@ def self.params_for_create
   def self.catalog_types
     {"autosde" => N_("Autosde")}
   end
+
+  def self.post_refresh_ems(ems_id, _)
+    events = EmsEvent.where("ems_id = ? AND (physical_storage_id IS NULL OR physical_storage_name IS NULL)", ems_id)
+
+    events.each do |e|
+      physical_storage_uuid = e.full_data[:storage_system]
+      physical_storage = PhysicalStorage.where(:ems_ref => physical_storage_uuid).to_a
```
agrare, line 305: If `physical_storage_uuid` is unique within the scope of an EMS then what you want here is to use `find_by` instead of `.where.first` (the only reason the bot didn't warn is because you did `.to_a` then `[0]` below)
```suggestion
      physical_storage_uuid = e.full_data[:storage_system]
      physical_storage = PhysicalStorage.find_by(:ems_id => ems_id, :ems_ref => physical_storage_uuid)
```

app/models/manageiq/providers/autosde/storage_manager.rb
```diff
@@ -296,4 +296,17 @@ def self.params_for_create
...
+
+      e.physical_storage_name = physical_storage[0].name
+      e.physical_storage_id = physical_storage[0].id
```
agrare, line 308: Then this becomes
```suggestion
      e.physical_storage_name = physical_storage.name
      e.physical_storage_id = physical_storage.id
```

Oferlis: Changed

app/models/manageiq/providers/autosde/storage_manager.rb
Oferlis: Changed

agrare (approved):

PR #200, opened 2023-01-29
storage service endpoint

OrGur1987:
see issue https://github.com/ManageIQ/manageiq-ui-classic/issues/8616

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/storage_service.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/storage_service.rb
@@ -1,2 +1,63 @@
 class ManageIQ::Providers::Autosde::StorageManager::StorageService < ::StorageService
+  supports :create
+  supports :delete
+
+  def self.raw_create_storage_service(ext_management_system, options = {})
+    creation_hash = {
+      :name                  => options['name'],
+      :description           => options['description'].nil? ? "none" : options['description'],
+      :capability_value_list => options['required_capabilities'].map { |capability| capability["value"] },
+      # currently only one default optional value is added in the backend for each of these,
+      # but in the future we might add them as miq models with more optional values:
+      :project               => "",
+      :profile               => "",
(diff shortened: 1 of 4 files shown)


OrGur1987:
https://codeclimate.com/github/ManageIQ/manageiq-providers-autosde/pull/200
these errors stem from several lines of dup code, can we ignore it?

app/models/manageiq/providers/autosde/storage_manager/storage_service.rb
```diff
@@ -1,2 +1,67 @@
 class ManageIQ::Providers::Autosde::StorageManager::StorageService < ::StorageService
+  supports :create
+  supports :delete
+  # supports :update  # will be added in the future
```
agrare, line 4: Best to just leave this out rather than leaving it commented out

app/models/manageiq/providers/autosde/storage_manager/storage_service.rb
```diff
@@ -1,2 +1,67 @@
...
+      options = {
+        :target_class   => :storage_services,
+        :target_id      => nil,
+        :ems_id         => ext_management_system.id,
+        :native_task_id => task_id,
+        :interval       => 10.seconds,
+        :target_option  => "new"
+      }
+      ext_management_system.class::EmsRefreshWorkflow.create_job(options).tap(&:signal_start)
+    end
```
agrare, line 31: Why is this in an ensure block?  You want to run the refresh workflow even if `ServiceApi.services_post` raises an exception?  But you're using the task_id so if that raised an exception task_id will be `nil`

OrGur1987: tbh I copied it from our physical_storage model, but it only appears there and not in any other raw_create in our provider

agrare (approved):
LGTM

PR #201, opened 2023-01-29
refresh bulk of volumes and get object_id as array

Lauferin:
When an object is created, it's auto-refreshed by the refresh worker once its status is success. But when a bulk of volumes is created, only one of the volumes was being refreshed because it got only one of the created volumes in the object_id.

I changed that to refresh all the ems (I would prefer it to refresh only cloud volumes or only those specific volumes but I don't think those options are supported).

Other than that, the object_id is now being received as a list, that's part of the changes of this PR. It's still in [wip] because the gem hasn't been updated yet.

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/ems_refresh_workflow.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/ems_refresh_workflow.rb
@@ -11,7 +11,7 @@ class ManageIQ::Providers::Autosde::StorageManager::EmsRefreshWorkflow < ManageI
     when "FAILURE"
       raise "#{options[:target_class]} task failed: #{JSON.parse(native_object.result)["exc_message"][0]}"
     when "SUCCESS"
-      options[:native_object_id] = native_object.object_id
+      options[:native_objects_ids] = native_object.objects_ids
       save!
       if options[:target_option] == 'new' && options[:target_class] == :physical_storages
         options[:target_option] = "ems"
@@ -31,7 +31,7 @@ class ManageIQ::Providers::Autosde::StorageManager::EmsRefreshWorkflow < ManageI
   def poll_refresh_task
     opts = {
       :query_params => {
-        :object_id => options[:native_object_id]
+        :object_id => options[:native_objects_ids][0]
       }
     }
     tasks = autosde_client.JobApi.jobs_get(opts)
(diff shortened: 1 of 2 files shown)


agrare:
@Lauferin you can do a targeted refresh of multiple objects, is it possible to get the ems_ref for each of the newly created volumes?

Lauferin:
Great! how?
yes, they're all in native_object.objects_ids

agrare:
You can simply pass an array of `InventoryRefresh::Target` objects so where you have this right now: https://github.com/ManageIQ/manageiq-providers-autosde/blob/master/app/models/manageiq/providers/autosde/storage_manager/ems_refresh_workflow.rb#L59-L63
You can do something like

```
targets = options[:native_object_id].to_a.map do |ems_ref|
  InventoryRefresh::Target.new(
    :manager     => ext_management_system,
    :association => options[:target_class],
    :manager_ref => {:ems_ref => ems_ref}
  )
end
```

And pass the whole thing to `EmsRefresh.queue_refresh_task(targets)`

Lauferin:
Great! thanks
This is still in wip because it's awaiting a gem update (which is itself awaiting approvals of other PRs)

Lauferin:
I removed the [wip] since the gem's version has been updated. Nothing to rebase.

app/models/manageiq/providers/autosde/storage_manager/ems_refresh_workflow.rb
```diff
@@ -31,7 +31,7 @@ def poll_native_task
   def poll_refresh_task
     opts = {
       :query_params => {
-        :object_id => options[:native_object_id]
+        :object_id => options[:native_objects_ids][0]
```
agrare, line 34: Hm this looks weird, do you want to wait until all volumes have finished instead of just the first one?

Lauferin: The method poll_refresh_task is only intended to deal with the case where we have to wait for a second task to success. This only happens when we add a physical storage (see lines 16 - 19) and then we have to wait for the autosde refresh (that's the second task) to succeed in order to do a general refresh.

In that case, there is only one object_id in the task.

If we add a bulk of volumes, all volumes will have finished their creation until the status of the task is success, in line 13.

Lauferin:
Sorry about the mess, I squashed into another commit that is not related. I'll try to fix this.

agrare (approved):

Lauferin:
Now it's ok :-)

PR #204, opened 2023-02-01
Gal/326036 add capabilities for create physical storage

galoiring:
adding the field for enabled capabilities in create physical storage.

Pr's:
- [ ] https://github.com/ManageIQ/manageiq-ui-classic/pull/8618

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
@@ -53,11 +53,12 @@ class ManageIQ::Providers::Autosde::StorageManager::PhysicalStorage < ::Physical
 
   def self.raw_create_physical_storage(ext_management_system, options = {})
     sys_to_create = ext_management_system.autosde_client.StorageSystemCreate(
-      :password       => options['password'],
-      :user           => options['user'],
-      :system_type    => PhysicalStorageFamily.find(options['physical_storage_family_id']).name,
-      :management_ip  => options['management_ip'],
-      :storage_family => "ontap_7mode"
+      :password                       => options['password'],
+      :user                           => options['user'],
+      :system_type                    => PhysicalStorageFamily.find(options['physical_storage_family_id']).name,
+      :management_ip                  => options['management_ip'],
+      :storage_family                 => "ontap_7mode",
+      :enabled_capability_values_list => options['enabled_capability_values']
     )
 
     begin
--- a/manageiq-providers-autosde.gemspec
+++ b/manageiq-providers-autosde.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_dependency "autosde_openapi_client", "~> 3.1.1"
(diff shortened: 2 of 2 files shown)


manageiq-providers-autosde.iml
```diff
@@ -0,0 +1,18 @@
...
+    <orderEntry type="library" scope="PROVIDED" name="ethon (v0.15.0, ruby-2.6.8-p205) [gem]" level="application" />
+    <orderEntry type="library" scope="PROVIDED" name="ffi (v1.15.5, ruby-2.6.8-p205) [gem]" level="application" />
+    <orderEntry type="library" scope="PROVIDED" name="typhoeus (v1.4.0, ruby-2.6.8-p205) [gem]" level="application" />
+  </component>
+  <component name="RakeTasksCache">
+    <option name="myRootTask">
+      <RakeTaskImpl id="rake" />
+    </option>
+  </component>
+</module>
```
agrare, line 18: Please add `*.iml` to your preferably your global `.gitignore` file so they aren't added automatically

app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
```diff
@@ -53,11 +53,12 @@ def raw_update_physical_storage(options = {})
 
   def self.raw_create_physical_storage(ext_management_system, options = {})
     sys_to_create = ext_management_system.autosde_client.StorageSystemCreate(
-      :password       => options['password'],
-      :user           => options['user'],
-      :system_type    => PhysicalStorageFamily.find(options['physical_storage_family_id']).name,
-      :management_ip  => options['management_ip'],
-      :storage_family => "ontap_7mode"
+      :password                       => options['password'],
```
agrare, line 56: If we aren't `ManageIQ::Password.encrypt`'ing the password here we need to since the `options` will end up in the queue and can be seen in the logs.  Then you will need to `ManageIQ::Password.try_decrypt(options['password'])` here

manageiq-providers-autosde.iml
galoiring: will do. sorry about that.

app/models/manageiq/providers/autosde/storage_manager/physical_storage.rb
galoiring: I will open a ticket for the next sprint to fix this and encrypt the password in all the relevant places

manageiq-providers-autosde.iml
agrare: You still need to delete this file

galoiring: done

agrare: Still still need to delete this file :wink:

galoiring: In each rebase this file come back to life 👻
I will delete it again

agrare: Recommend rebasing the original commit bb7058bd72e4845b5eadfaf22a295d7b83b7dae1 and removing the file from there so we aren't tracking an add+remove of a file we don't want

galoiring:
@agrare is this ready to be merge ? 
(just to make sure, so that after Jeff will assist on the ui pr we will be ready to do the merge. )

agrare (approved):

PR #205, opened 2023-02-08
Add system type to the asset detail for textual summary

galoiring:
We added the physical storage family (the storage model) to the correct place.

Screenshot 2023-02-08 at 11 33 42

### DIFF
--- a/app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
+++ b/app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
@@ -45,12 +45,13 @@ class ManageIQ::Providers::Autosde::Inventory::Parser::StorageManager < ManageIQ
 
   def physical_storages
     collector.physical_storages.each do |storage|
-      persister.physical_storages.build(
+      physical_storage = persister.physical_storages.build(
         :name                    => storage.name,
         :ems_ref                 => storage.uuid,
         :physical_storage_family => persister.physical_storage_families.lazy_find(storage.system_type.uuid),
         :health_state            => storage.status
       )
+      persister.physical_storage_details.build(:resource => physical_storage, :model => storage.system_type.name)
     end
   end
 
--- a/app/models/manageiq/providers/autosde/inventory/persister/storage_manager.rb
+++ b/app/models/manageiq/providers/autosde/inventory/persister/storage_manager.rb
@@ -18,5 +18,16 @@ class ManageIQ::Providers::Autosde::Inventory::Persister::StorageManager < Manag
     add_collection(storage, :cloud_volumes)
     add_collection(storage, :ext_management_system)
     add_collection(storage, :storage_service_resource_attachments)
+    add_physical_storage_details
+  end
+
+  def add_physical_storage_details
+    add_collection(storage, :physical_storage_details) do |builder|
+      builder.add_properties(
+        :model_class                  => ::AssetDetail,
+        :manager_ref                  => %i[resource],
+        :parent_inventory_collections => %i[physical_storages]
+      )
+    end
   end
 end


app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
```diff
@@ -48,7 +48,8 @@ def physical_storages
         :name                    => storage.name,
         :ems_ref                 => storage.uuid,
         :physical_storage_family => persister.physical_storage_families.lazy_find(storage.system_type.uuid),
-        :health_state            => storage.status
+        :health_state            => storage.status,
+        :asset_detail            => AssetDetail.new(:model => storage.system_type.name)
```
agrare, line 52: You don't want to build an `AssetDetail` class instance here what you want is to use the `physical_storage_details` inventory collection:

```suggestion
        :asset_detail            => persister.physical_storage_details.build(:model => storage.system_type.name)
```

(You'll probably need to add this to your inventory-collections set https://github.com/ManageIQ/manageiq-providers-autosde/blob/master/app/models/manageiq/providers/autosde/inventory/persister/storage_manager.rb#L8-L20)

agrare:
```
+    add_physical_storage_details
     add_collection(storage, :physical_storage_families)
     add_collection(storage, :storage_resources)
     add_collection(storage, :storage_services)
@@ -19,4 +20,14 @@ class ManageIQ::Providers::Autosde::Inventory::Persister::StorageManager < Manag
     add_collection(storage, :ext_management_system)
     add_collection(storage, :storage_service_resource_attachments)
   end
+
+  def add_physical_storage_details
+    add_collection(storage, :physical_storage_details) do |builder|
+      builder.add_properties(
+        :model_class                  => ::AssetDetail,
+        :manager_ref                  => %i[resource],
+        :parent_inventory_collections => %i[:physical_storage]
+      )
+    end
+  end
```

galoiring: > You don't want to build an `AssetDetail` class instance here what you want is to use the `physical_storage_details` inventory collection:
> 
> ```
>         physical_storage = persister.physical_storages.build(...)
>         persister.physical_storage_details.build(:resource => physical_storage, :model => storage.system_type.name)
> ```
> 
> (You'll probably need to add this to your inventory-collections set https://github.com/ManageIQ/manageiq-providers-autosde/blob/master/app/models/manageiq/providers/autosde/inventory/persister/storage_manager.rb#L8-L20)

and I need this code snippet to be added to asset_detail inside physical storage ? (line 52)
doing so and the ui will recognize it and will present it at the right places.

agrare:
Looks good you just have to add the relation `has_many :physical_storage_details, :through => :physical_storages, :source => :asset_detail` to your StorageManager class

galoiring:
> 

This line already exist, line 35 in storage manager.

agrare:
Okay your specs are failing for a different reason then.

galoiring:
> Okay your specs are failing for a different reason then.

I will try to check it out. 
this is the correct way to add this asset detail ?
I don't need to call this 'add_physical_storage_details' function inside the 'parser/storage_manager.rb' ?

galoiring:
@agrare ok I fix it and now it work as you suggested.
I didn't make the spec to pass,
try to debug it : 
Screenshot 2023-03-19 at 15 19 32

but specific in the refresher spec line 68, the 'physical_storage_details' isn't recognize properly : 

Screenshot 2023-03-19 at 15 18 56

app/models/manageiq/providers/autosde/inventory/persister/storage_manager.rb
```diff
@@ -19,4 +19,14 @@ def initialize_inventory_collections
     add_collection(storage, :ext_management_system)
     add_collection(storage, :storage_service_resource_attachments)
   end
+
+  def physical_storage_details
+    add_collection(storage, :physical_storage_details) do |builder|
+      builder.add_properties(
+        :model_class                  => ::AssetDetail,
+        :manager_ref                  => %i[resource],
+        :parent_inventory_collections => %i[:physical_storage]
```
agrare, line 28:
```suggestion
        :parent_inventory_collections => %i[physical_storages]
```

app/models/manageiq/providers/autosde/inventory/persister/storage_manager.rb
```diff
@@ -19,4 +19,14 @@ def initialize_inventory_collections
     add_collection(storage, :ext_management_system)
     add_collection(storage, :storage_service_resource_attachments)
   end
+
+  def physical_storage_details
```
agrare, line 23: You're actually overwriting the accessor method `persister.physical_storage_details` that is being used by the parser so each time you call `persister.physical_storage_details.build` it is calling this method.

```suggestion
    add_physical_storage_details
  end

  def add_physical_storage_details
```

agrare:
Just need to fix the rubocop warning then LGTM

galoiring:
> Just need to fix the rubocop warning then LGTM

Thanks a lot !

galoiring:
> Just need to fix the rubocop warning then LGTM

@agrare thank you for your help! I does pass the specs.

PR #213, opened 2023-02-19
Capabilities fields modifiaction and parsing

galoiring:
This Pr is about the  jsonb column for capabilities as we discussed.

followup to this pr: https://github.com/ManageIQ/manageiq-schema/pull/677

UI: https://github.com/ManageIQ/manageiq-ui-classic/pull/8656

w/ @OrGur1987

### DIFF
--- a/app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
+++ b/app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
@@ -28,7 +28,7 @@ class ManageIQ::Providers::Autosde::Inventory::Parser::StorageManager < ManageIQ
   def ext_management_system
     persister.ext_management_system.build(
       :guid         => persister.manager.guid,
-      :capabilities => collector.capability_values
+      :capabilities => parse_ems_capabilities(collector.capability_values)
     )
   end
 
@@ -38,7 +38,7 @@ class ManageIQ::Providers::Autosde::Inventory::Parser::StorageManager < ManageIQ
         :name         => storage_family.name,
         :ems_ref      => storage_family.uuid,
         :version      => storage_family.version,
-        :capabilities => parse_capabilities(storage_family.capability_values_json, 'abstract_capability__name')
+        :capabilities => parse_possible_capabilities(storage_family.capability_values_json, 'abstract_capability__name')
       )
     end
   end
@@ -62,7 +62,7 @@ class ManageIQ::Providers::Autosde::Inventory::Parser::StorageManager < ManageIQ
         :logical_free     => resource.logical_free,
         :logical_total    => resource.logical_total,
         :physical_storage => persister.physical_storages.lazy_find(resource.storage_system),
-        :capabilities     => parse_capabilities(resource.capability_values_json, 'abstract_capability')
+        :capabilities     => parse_possible_capabilities(resource.capability_values_json, 'abstract_capability')
       )
     end
   end
@@ -130,7 +130,7 @@ class ManageIQ::Providers::Autosde::Inventory::Parser::StorageManager < ManageIQ
         :description  => service.description,
         :version      => service.version,
         :ems_ref      => service.uuid,
-        :capabilities => parse_capabilities(service.capability_values_json, 'abstract_capability')
+        :capabilities => parse_service_capabilities(service.capability_values_json, 'abstract_capability')
       )
     end
   end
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
```diff
@@ -181,11 +181,35 @@ def storage_service_resource_attachments
     end
   end
 
-  # This method changes capability name field from 'abstract_capability' to 'name'
-  def parse_capabilities(capabilities, field_name)
+  def parse_ems_capabilities(capabilities)
+    return capabilities unless capabilities.is_a?(Array)
```
agrare, line 185: Is this trying to catch when capabilities is `nil` ?  Or can capabilities be other types like a Hash?
If the former then best to just `return if capabilities.nil?`  If the later what does the format of the hash look like?

app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
```diff
@@ -181,11 +181,35 @@ def storage_service_resource_attachments
...
+
+    caps_hash = {}
+    capabilities.each do |capability|
+      name = capability.abstract_capability
+      (caps_hash[name] ||= []) << {"uuid" => capability.uuid, "value" => capability.value}
+    end
+    caps_hash
```
agrare, line 192:
```suggestion
    capabilities.to_h do |capability|
      [capability.abstract_capability, {"uuid" => capability.uuid, "value" => capability.value}]
    end
```

app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
```diff
@@ -181,11 +181,35 @@ def storage_service_resource_attachments
...
+  end
+
+  def parse_possible_capabilities(capabilities, field_name)
+    caps = JSON.parse(capabilities)
+    caps_hash = {}
+    caps.each do |capability|
+      name = capability[field_name]
+      (caps_hash[name] ||= []) << capability["value"]
+    end
+    caps_hash
```
agrare, line 202:
```suggestion
    JSON.parse(capabilities).to_h do |capability|
      [capability[field_name], capability["value"]]
    end
```

app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
```diff
@@ -181,11 +181,35 @@ def storage_service_resource_attachments
...
+
+  def parse_service_capabilities(capabilities, field_name)
     caps = JSON.parse(capabilities)
+    caps_hash = {}
     caps.each do |capability|
-      capability['name'] = capability.delete(field_name)
+      name = capability[field_name]
+      caps_hash[name] = capability["value"]
     end
+    caps_hash
```
agrare, line 212: Similar suggestion using to_h over a local-variable just to set in the loop then return it

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
```diff
@@ -110,8 +110,11 @@ def params_for_update
   end
 
   def self.params_for_create(provider)
-    capabilities = provider.capabilities.map do |capability|
-      {:label => "#{capability['abstract_capability']}: #{capability['value']}", :value => capability['uuid']}
+    capabilities = provider.capabilities.flat_map do |capability|
+      capability_name = capability.first
+      capability.last.map do |c|
```
agrare, line 115: When you loop over a hash you can do:
```
{:a => "a", :b => "b"}.each do |key, val|
```
so that you don't have to do `capability.first` and `capability.last`

```suggestion
    capabilities = provider.capabilities.flat_map do |name, capability|
      capability.map do |c|
```

app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
agrare: Just noticed you expect to have >1 value per key.  You can use `each_with_object` instead and you can use `Hash.new { |k, v| k[v] = []}` so you don't need the `(caps_hash[name] ||= [])`

Try `capabilities.each_with_object(Hash.new { |k, v| k[v] = []}) do |capability, result|`

app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
agrare: Still have a question here

OrGur1987: > Is this trying to catch when capabilities is `nil` ? Or can capabilities be other types like a Hash?
> If the former then best to just `return if capabilities.nil?` If the later what does the format of the hash look like?

Hi, so the parameter `capabilities` we receive here is `collector.capability_values` which holds either the array of capabilities we received from the backend, or the already populated hash of capabilities we save in the provider.

If it is the former, we would like to parse it and populate the latter. If it's already the latter, then we would like to keep it the way it is.

agrare: Interesting feel like there's a better way to handle this but we don't have to fix that here

app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
```diff
@@ -181,11 +181,30 @@ def storage_service_resource_attachments
...
+    end
+  end
+
+  def parse_possible_capabilities(capabilities, field_name)
+    caps_hash = {}
+    JSON.parse(capabilities).each do |capability|
+      name = capability[field_name]
+      (caps_hash[name] ||= []) << capability["value"]
     end
+    caps_hash
```
agrare, line 198: Again style wise we try not to do
```
result = {}
loop do
  result << something
end
result
```

Better to use `.to_h` or `.each_with_object` depending on your use case.  Can you tackle this in a follow-up immediately after this?

agrare (approved):

app/models/manageiq/providers/autosde/inventory/parser/storage_manager.rb
OrGur1987: if u happen to come up with a better way, we could include it in the fix you asked for in the comment below.

PR #214, opened 2023-02-23
select storage resources to attach to in storage service create

OrGur1987:
- storage service: ems refresh for service create and delete is general and not targeted, so that service-resource-attachments will be refreshed as well.

- cloud volume: create performs a general refresh if:
  - mode was advanced: so that the new service and its attachments to the selected resources will be refreshed as well.
  - count > 1: so that all the new volumes will be refreshed

- [ ] https://github.com/ManageIQ/manageiq-ui-classic/pull/8662

creating a storage with attached resources now enables to select it in volume create basic mode:
image

and the volumes will be created on one of its attached resources:
image

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
@@ -19,16 +19,20 @@ class ManageIQ::Providers::Autosde::StorageManager::CloudVolume < ::CloudVolume
 
     if options['mode'] == 'Basic'
       creation_hash[:service] = ext_management_system.storage_services.find(options["storage_service_id"]).ems_ref
+      target_class = :cloud_volumes
+      target_option = "new"
     else
       creation_hash[:service_name] = options["new_service_name"]
       creation_hash[:resources] = ext_management_system.storage_resources.find(options["storage_resource_id"].to_a.pluck("value")).pluck(:ems_ref)
       creation_hash[:service_capabilities] = options['required_capabilities'].map { |capability| capability["value"] }
+      target_class = nil
+      target_option = "ems"
     end
 
     vol_to_create = ext_management_system.autosde_client.VolumeCreate(creation_hash)
     task_id = ext_management_system.autosde_client.VolumeApi.volumes_post(vol_to_create).task_id
 
-    create_refresh_task(nil, task_id, :cloud_volumes, ext_management_system, "new")
+    create_refresh_task(nil, task_id, target_class, ext_management_system, target_option)
   end
 
   # ================= delete  ================
@@ -140,7 +144,7 @@ class ManageIQ::Providers::Autosde::StorageManager::CloudVolume < ::CloudVolume
           :component    => "radio",
           :name         => "mode",
           :id           => "mode",
-          :label        => _("Mode"),
+          :label        => _("Mode: Create volume(s) using an existing storage service or create a new service as well"),
           :initialValue => 'Basic',
           :options      => [{:label => 'Basic', :value => 'Basic',}, {:label => 'Advanced', :value => 'Advanced'}],
           :isRequired   => true,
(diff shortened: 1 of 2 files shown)


OrGur1987:
@agrare :)

OrGur1987:
@agrare 
https://github.com/ManageIQ/manageiq-ui-classic/pull/8662 approved by @jeffibm

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
```diff
@@ -19,16 +19,25 @@ def self.raw_create_volume(ext_management_system, options = {})
 
     if options['mode'] == 'Basic'
       creation_hash[:service] = ext_management_system.storage_services.find(options["storage_service_id"]).ems_ref
+      if options["count"] == 0
+        target_class = :cloud_volumes
+        target_option = "new"
+      else
+        target_class = nil
+        target_option = "ems"
```
agrare, line 27: Follow-up, but we can do targeted refresh of multiple targets which will still be way more efficient than a full ems refresh.  Can you get the list of ems_refs of the new volumes?

app/models/manageiq/providers/autosde/storage_manager/storage_service.rb
```diff
@@ -17,25 +18,25 @@ def self.raw_create_storage_service(ext_management_system, options = {})
...
     task_id = ext_management_system.autosde_client.ServiceApi.services_post(service_to_create).task_id
     options = {
-      :target_class   => :storage_services,
+      :target_class   => nil,
       :target_id      => nil,
       :ems_id         => ext_management_system.id,
       :native_task_id => task_id,
       :interval       => 10.seconds,
-      :target_option  => "new"
+      :target_option  => "ems"
```
agrare, line 26: Why this change?  Is the targeted refresh not working?  Seems like something we should fix not just do a full.

app/models/manageiq/providers/autosde/storage_manager/storage_service.rb
```diff
@@ -17,25 +18,25 @@ def self.raw_create_storage_service(ext_management_system, options = {})
...
     options = {
-      :target_id      => id,
-      :target_class   => self.class.name,
+      :target_id      => nil,
+      :target_class   => nil,
       :ems_id         => ext_management_system.id,
       :native_task_id => task_id,
       :interval       => 20.seconds,
-      :target_option  => "existing"
+      :target_option  => "ems"
```
agrare, line 39: Same here, this looks unrelated to the rest of the changes and delete should be a targeted refresh.

OrGur1987: I've opened #220 for this particular change and will remove it from this PR

app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
OrGur1987: the advanced volume create leads to the backend first creating a new service, attaching it to the selected resources and creating the volume(s) on one of the resources. so we get the `ems_ref`s of the new volume(s), but not that of the new service and `storage_service_resource_attachments`.

if triggering a full refresh is too much of an overkill, we'll leave it as targeted for what we can get, and we'll just have to manually perform a full refresh until we manage to get `ems_ref`s for various types of objects in one job.

OrGur1987:
@agrare following your comments, the current version has a targeted refresh on service create and delete and volume create basic, and a full one only for volume create advanced. if you prefer a targeted refresh on volume create advanced as well, I'll change it, and we'll have to manually refresh until we manage to get `ems_ref`s for various types of objects in one job.

agrare:
@OrGur1987 no issue with handling the advanced case in a follow-up

agrare (approved):

PR #219, opened 2023-03-12
split volume create capabilities to separate single select fields

OrGur1987:
`params_for_create` now generates a single-select field for each capability for filtering services/resources

pr in ui repo:
- [ ] https://github.com/ManageIQ/manageiq-ui-classic/pull/8702


app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
```diff
@@ -109,9 +109,20 @@ def params_for_update
 
   def self.params_for_create(provider)
     capabilities = provider.capabilities.flat_map do |name, values|
-      values.map do |c|
-        {:label => "#{name}: #{c['value']}", :value => c['uuid']}
-      end
+      [
```
agrare, line 112: Why `flat_map` an array of one item?  Can't you just `.map` the single hash?
```
capabilities = provider.capabilities.flat_map do |name, values|
  {
    :component ...
```

agrare (approved):

PR #220, opened 2023-03-13
refresh after delete service

OrGur1987:
added `StorageService` to `TargetCollection#parse_targets!`.
without that, a post-service_delete targeted refresh reaches `TargetCollection#storage_services` without an ems_ref.

- [x] https://github.com/ManageIQ/manageiq/pull/22404

### DIFF
--- a/app/models/manageiq/providers/autosde/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/autosde/inventory/collector/target_collection.rb
@@ -78,6 +78,8 @@ class ManageIQ::Providers::Autosde::Inventory::Collector::TargetCollection < Man
         add_target!(:physical_storages, target.ems_ref)
       when CloudVolume
         add_target!(:cloud_volumes, target.ems_ref)
+      when StorageService
+        add_target!(:storage_services, target.ems_ref)
       end
     end
   end
--- a/spec/models/manageiq/providers/autosde/storage_manager/refresher_spec.rb
+++ b/spec/models/manageiq/providers/autosde/storage_manager/refresher_spec.rb
@@ -36,6 +36,7 @@ describe ManageIQ::Providers::Autosde::StorageManager::Refresher do
       let(:system_type_api)    { double("SystemTypeApi") }
       let(:storage_system_api) { double("StorageSystemApi") }
       let(:volume_api)         { double("VolumeApi") }
+      let(:service_api)        { double("ServiceApi") }
 
       it "with no targets" do
         assert_inventory_not_changed { run_targeted_refresh }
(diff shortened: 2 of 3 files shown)


agrare:
@OrGur1987 can you write a spec that demonstrates the issue?

OrGur1987:
> @OrGur1987 can you write a spec that demonstrates the issue?

tbh I'm not sure how.. 

I tried adding an `it "deleting a storage service"` to `spec/models/manageiq/providers/autosde/storage_manager/refresher_spec.rb` but these specs explicitly call a targeted refresh with a `manager_ref` argument, which isn't exactly how it actually works in `ManageIQ::Providers::Autosde::StorageManager::EmsRefreshWorkflow#refresh`  

I also tried adding a `get(api_storage_services_url)` to `it "deletes a single Storage Service"` in `spec/requests/storage_services_spec.rb` and to count the size of `response.parsed_body["resources"]`.
it remains on size 1 after the delete, but the same happens when I do it in `it "Deletion of a single Physical Storage"`, even though I know that a targeted refresh after deleting a physical_storage does succeed.  

I've added `StorageService` to `EmsRefresh.get_target_objects` here:
https://github.com/ManageIQ/manageiq/pull/22404
before that it didn't even reach `ManageIQ::Providers::Autosde::Inventory::Collector::TargetCollection#storage_services`

but it gets there with an empty `target.manager_refs_by_association`, in contrast to `physical_storage` for instance.

agrare:
```
adam@desktop:~/src/manageiq/manageiq-providers-autosde$ git diff
diff --git a/spec/models/manageiq/providers/autosde/storage_manager/refresher_spec.rb b/spec/models/manageiq/providers/autosde/storage_manager/refresher_spec.rb
index b5abc6a..9b36553 100644
--- a/spec/models/manageiq/providers/autosde/storage_manager/refresher_spec.rb
+++ b/spec/models/manageiq/providers/autosde/storage_manager/refresher_spec.rb
@@ -33,6 +33,7 @@ describe ManageIQ::Providers::Autosde::StorageManager::Refresher do
     context "targeted refresh" do
       before { VCR.use_cassette("ems_refresh_v1") { described_class.refresh([ems]) } }
 
+      let(:service_api)        { double("ServiceApi") }
       let(:system_type_api)    { double("SystemTypeApi") }
       let(:storage_system_api) { double("StorageSystemApi") }
       let(:volume_api)         { double("VolumeApi") }
@@ -185,6 +186,17 @@ describe ManageIQ::Providers::Autosde::StorageManager::Refresher do
         expect(ems.cloud_volumes.count).to(eq(12))
       end
 
+      it "deleting a storage service" do
+        expect(service_api).to(receive(:services_get).and_return([]))
+
+        storage_service = ems.storage_services.find_by(:ems_ref => "774c1fd8-43e6-4bb2-8466-d5d1c1d992d6")
+        run_targeted_refresh(InventoryRefresh::Target.new(:manager => ems, :association => :storage_services, :manager_ref => {:ems_ref => storage_service.ems_ref}))
+
+        ems.reload
+
+        expect(ems.storage_services.count).to(eq(6))
+      end
+
       def run_targeted_refresh(targets = [])
         target = InventoryRefresh::TargetCollection.new(:manager => ems, :targets => Array(targets))
 
@@ -194,6 +206,7 @@ describe ManageIQ::Providers::Autosde::StorageManager::Refresher do
 
         # Allow for tests to mock API calls
         autosde_client_stub = double("AutosdeClient")
+        allow(autosde_client_stub).to(receive(:ServiceApi).and_return(service_api))
         allow(autosde_client_stub).to(receive(:SystemTypeApi).and_return(system_type_api))
         allow(autosde_client_stub).to(receive(:StorageSystemApi).and_return(storage_system_api))
         allow(autosde_client_stub).to(receive(:VolumeApi).and_return(volume_api))
```

You're right in that you can't pass a `StorageService` object in currently, you also have to handle this in the `parse_targets!` method in your TargetCollector.

Passing in an active_record object like that is just short-hand for passing in a InventoryRefresh::Target so you have to transform it there.

agrare:
@OrGur1987 can you add the spec above to ensure this works as expected?

agrare (approved):

PR #222, opened 2023-03-22
[wip] change iqn and wwpn join char to pipe

llivne:


agrare:
Wonder if a comma separated list or a JSON array would be better

llivne:
> Wonder if a comma separated list or a JSON array would be better

@agrare  autosde is using the same logic like svc storage which is a comma-separated list (string) of addresses:
https://www.ibm.com/docs/en/flashsystem-9x00/8.3.x?topic=commands-mkhost

I also thinking json array may be a good option, but in this case I think it would be better to use the same format like the storages.

llivne:
> @llivne this is being sent to the AutoSDE API right? Is a change necessary on that side?

yes, this is sent to AutoSDE API. no changes needed in AutoSDE side.

agrare:
> yes, this is sent to AutoSDE API. no changes needed in AutoSDE side.

How can we change the separator without changing what is processing the separator as well?

llivne:
> > yes, this is sent to AutoSDE API. no changes needed in AutoSDE side.
> 
> How can we change the separator without changing what is processing the separator as well?

@agrare you are correct, I also changed the delimiter type in autosde since we used a wrong type that can be part of the address name. this was a bug in both miq autosde provider and autosde. what I meant is that there are no changes in the autosde gem.

PR #226, opened 2023-04-10
storage service edit mode: check which attached resources are compliant to selected capabilities

OrGur1987:
issue: https://github.com/ManageIQ/manageiq-ui-classic/issues/8749

added button in service edit mode which enables user to check what currently attached resources will comply with the selected capabilities.

tis PR adds the necessary model methods to support this new action.


OrGur1987:
@agrare ready for review, thanks

app/models/manageiq/providers/autosde/storage_manager/storage_service.rb
```diff
@@ -46,16 +47,32 @@ def raw_delete_storage_service
...
+    current_capabilities = ext_management_system.capabilities.slice(*capabilities.keys)
+    current_capabilities_refs = []
+    capabilities.each do |key, value|
+      current_capabilities[key].each do |cap|
+        current_capabilities_refs.push(cap["uuid"]) if cap["value"] == value
+      end
+    end
+    capability_value_list = if current_capabilities_refs.sort != selected_capabilities_refs.sort
+                              selected_capabilities_refs
+                            end
```
agrare, line 68: @OrGur1987 this is complex enough now to require a spec test demonstrating the various cases that you are handling here

app/models/manageiq/providers/autosde/storage_manager/storage_service.rb
```diff
@@ -46,16 +47,32 @@ def raw_delete_storage_service
...
 
     update_details = ext_management_system.autosde_client.ServiceUpdate(
       :name               => options['name'],
       :description        => options['description'],
-      :resources          => ext_management_system.storage_resources.find(options["storage_resource_id"].to_a.pluck("value")).pluck(:ems_ref),
-      :capability_id_list => capability_value_list
+      :resources          => resources || nil,
+      :capability_id_list => capability_value_list || nil
```
agrare, line 74: `resources || nil` doesn't make sense (unless resources could be `false`).  If `resources` is `nil` then `nil || nil` is not useful.

app/models/manageiq/providers/autosde/storage_manager/storage_service.rb
```diff
@@ -46,16 +47,32 @@ def raw_delete_storage_service
...
-    capability_value_list.delete("-1")
+    # send resources only if edited
+    current_resources = storage_resources.ids
+    selected_resources = options["storage_resource_id"].to_a.pluck("value").map(&:to_i)
+    resources = if current_resources.sort != selected_resources.sort
+                  ext_management_system.storage_resources.find(options["storage_resource_id"].to_a.pluck("value")).pluck(:ems_ref)
+                end
+    # send capability_value_list only if edited
+    selected_capabilities_refs = options.slice(*ext_management_system.capabilities.keys).values
+    selected_capabilities_refs.delete("-1")  # get rid of the N/A option
```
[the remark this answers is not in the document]
OrGur1987: `compression` and `thin_provision` are in `ext_management_system.capabilities.keys` and their value might be '-1' if the user selected 'N/A'. we want to filter this out before checking equality against `current_capabilities`.

```ruby
[68, 77] in /var/www/miq/vmdb/plugins/manageiq-providers-autosde/app/models/manageiq/providers/autosde/storage_manager/storage_service.rb
   68:                             end
   69: 
   70:     require 'byebug'
   71:     byebug
   72: 
=> 73:     update_details = ext_management_system.autosde_client.ServiceUpdate(
   74:       :name               => options['name'],
   75:       :description        => options['description'],
   76:       :resources          => resources || nil,
   77:       :capability_id_list => capability_value_list || nil
(byebug) ext_management_system.capabilities.keys
["compression", "data_reduction", "thin_provision"]
(byebug) options.slice(*ext_management_system.capabilities.keys).values
["-1", "9ae86a32-e64a-4d0a-9d8e-ece05ad7c401"]
(byebug) 

```

app/models/manageiq/providers/autosde/storage_manager/storage_service.rb
OrGur1987: u mean a spec that tests the creation hash received by `ext_management_system.class::EmsRefreshWorkflow.create_job()` at the end of this method in the various cases? 

can I mock `EmsRefreshWorkflow` and check what it receives from `raw_update_storage_service()`?

agrare: More importantly what `services_pk_put` is receiving based on the various scenarios on current capabilities vs selected capabilities

OrGur1987: how can I mock/check that?
we don't have too many examples in our provider repo spec dir
can u perhaps point me to a useful example from another provider?
thanks

OrGur1987: hi @agrare, 
@jeffibm merged the ui for this issue

OrGur1987: @agrare reminder
thanks

agrare: @OrGur1987 you are looking for examples of spec tests?  You would return a `double("autosde_client")` (https://www.rubyguides.com/2018/10/rspec-mocks/) so you would `expect(ems).to receive(:autosde_client).and_return(autosde_client)`
Then you can `expect(autosde_client).to receive(:ServiceUpdate).with(expected_args).and_return(something)`

OrGur1987: I thought it might be easier and cleaner to export the preprocessing of the selected resources and capabilities to a helper method (say `prepare_update_args`), and then only test that in the spec, since there's no need to actually test the `client`.

something like:
```ruby
  let(:ems) do
    FactoryBot.create(
      # copied from some example, not sure it's needed here:
      :autosde_storage_manager,
      :with_autosde_credentials,
      :hostname     => Rails.application.secrets.autosde[:appliance_host],
      
      # dummy capabilities:
      :capabilities => {
        "cap1": [
          {
            "uuid": "111",
            "value": "True"
          },
          {
            "uuid": "222",
            "value": "False"
          }
        ],
        "cap2": [
          {
            "uuid": "333",
            "value": "True"
          },
          {
            "uuid": "444",
            "value": "False"
          }
        ],
      }
    )
    end

    let(:test_resource1) do
      FactoryBot.create(:storage_resource, :ext_management_system => ems)
    end

    let(:test_resource2) do
      FactoryBot.create(:storage_resource, :ext_management_system => ems)
    end

    let(:test_service) do
      FactoryBot.create(
        :storage_service,
        :ext_management_system => ems,
        :storage_resources     => [test_resource1, test_resource2],
        :capabilities          => {"cap1": "True", "cap2": "False"}
        )
    end

it "calls prepare update with the existing resources and capabilities" do
  test_service.prepare_update_args(options = {existing_resources_and_capabilities})
  ...
  ...
end
``` 

but the created `test_service` seems to be `ManageIQ::StorageService` and not `StorageManager::StorageService`, and I can't access a `StorageManager::StorageService` method without adding it to the super as well..

is there a way to `FactoryBot.create` the `StorageManager::StorageService`?

OrGur1987: found out that specifying the type in the create args produces the desired child-model.
is the added spec ok with u?
I did find a tiny bug in the ui and will open another pr for it

spec/models/manageiq/providers/autosde/storage_service_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  before(:each) do
+    @test_service = FactoryBot.create(
+      :storage_service,
+      :type                  => "ManageIQ::Providers::Autosde::StorageManager::StorageService",
+      :ext_management_system => ems,
+      :storage_resources     => [test_resource1, test_resource2],
+      :capabilities          => {'cap1' => "True", 'cap2' => "False"}
+    )
+  end
```
agrare, line 34: Minor but it is preferable to use `let(:test_service)` (or `let!` if you need the object to be created prior to being referenced)
```suggestion
  let!(:test_service) do
    FactoryBot.create(
      :storage_service,
      :type                  => "ManageIQ::Providers::Autosde::StorageManager::StorageService",
      :ext_management_system => ems,
      :storage_resources     => [test_resource1, test_resource2],
      :capabilities          => {'cap1' => "True", 'cap2' => "False"}
    )
  end
```

agrare (approved):
It is a little weird that we're testing the prepare_options_hash method but I get that you were having issues setting the spec tests up.  I don't want to hold this up for that I'll try to throw an example up if I can get some time.

PR #227, opened 2023-04-16
730715 custom task queuing massage

galoiring:
currently miq task queue shows a success message for sending a request to the backend as if the backend actually succeeded in performing the request, even if the request subsequently fails:
60eca5a2-9295-43dc-a38e-35e757503886

we changed this inside task_opts so that the action label will show queuing massage correctly:
Screenshot 2023-04-17 at 13 44 59

Related code pr: https://github.com/ManageIQ/manageiq/pull/22459

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/cloud_volume.rb
@@ -40,6 +40,16 @@ class ManageIQ::Providers::Autosde::StorageManager::CloudVolume < ::CloudVolume
     create_refresh_task(nil, task_id, target_class, ext_management_system, target_option)
   end
 
+  def self.create_volume_queue(userid, ext_management_system, options = {})
+    task_opts = {
+      :action => "Queuing create Cloud Volume for user #{userid}",
+      :userid => userid
+    }
+
+    super(userid, ext_management_system, options, task_opts)
+  end
+
+
   # ================= delete  ================
 
   def raw_delete_volume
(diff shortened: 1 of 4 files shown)


agrare:
@galoiring what I'd rather do is to hook your ems_refresh_workflow to the original task so that it doesn't show complete just for delivering to the backend.

galoiring:
> @galoiring what I'd rather do is to hook your ems_refresh_workflow to the original task so that it doesn't show complete just for delivering to the backend.

but it is polling the task itself until 'success' and after that signal to the refresh task to initiate.
(the queuing is just for the task itself - create volume in this screenshot)

galoiring:
as you can see here in line 40, it will create the refresh task. 

Screenshot 2023-04-18 at 16 11 18

agrare:
Sorry I'm having a hard time following.

It sounds like we're marking the `"creating Cloud Volume for user #{userid}"` task complete+successful too soon if the cloud_volume creation actually fails.

Which task is failing?

galoiring:
> Sorry I'm having a hard time following.
> 
> It sounds like we're marking the `"creating Cloud Volume for user #{userid}"` task complete+successful too soon if the cloud_volume creation actually fails.
> 
> Which task is failing?
> 
> Can we re-use the initial task to track whatever work the second task is doing?

the create cloud volume massage is too soon yes , 
we need to mark this task as queued (if possible), and mark it as success just after it finished (and after that it completed successfully, the emsrefresh will initiate)

agrare:
> the create cloud volume massage is too soon yes ,

Okay then I think instead of changing the message here we should be re-using the same task to track the entire creation process.

PR #228, opened 2023-04-18
Fixed massage emsrefresh completed successfully

galoiring:
currently there's no dedicated success message for propagating success from autosde backend to miq_task.
we only get an EmsRefresh success msg like:

a93b3af1-f7cb-4a64-ad4d-3b5b87161e8c

After the changes we made it looks like this : 
Screenshot 2023-04-18 at 15 43 35

Related PR: 

- [ ] Core: https://github.com/ManageIQ/manageiq/pull/22465

### DIFF
--- a/app/models/manageiq/providers/autosde/storage_manager/ems_refresh_workflow.rb
+++ b/app/models/manageiq/providers/autosde/storage_manager/ems_refresh_workflow.rb
@@ -70,7 +70,7 @@ class ManageIQ::Providers::Autosde::StorageManager::EmsRefreshWorkflow < ManageI
     else
       signal(:abort, "error, no valid option")
     end
-    task_ids = EmsRefresh.queue_refresh_task(targets)
+    task_ids = queue_refresh_task(targets)
     if task_ids.blank?
       process_error("Failed to queue refresh", "error")
       queue_signal(:error)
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/autosde/storage_manager/ems_refresh_workflow.rb
```diff
@@ -80,8 +80,20 @@ def refresh
...
+    task_options = {
+      :action => "EmsRefresh completed successfully ",
+      :userid => "system"
+    }
+
+    EmsRefresh.queue_refresh_task(target, id, task_options)
+  end
+
+
+
```
Fryguy, line 96: Minor style, but can you trim these multiple blank lines to a single blank line?

agrare:
It seems weird that the "action" would be "completed successfully"

Also this isn't autosde specific so if we're going to change the task probably should just be done here https://github.com/ManageIQ/manageiq/blob/master/app/models/ems_refresh.rb#L169-L172

galoiring:
> It seems weird that the "action" would be "completed successfully", it does look like we could improve the task message though the targets aren't very helpful when they aren't class+id pairs.
> 
> Also this isn't autosde specific so if we're going to change the task probably should just be done here https://github.com/ManageIQ/manageiq/blob/master/app/models/ems_refresh.rb#L169-L172

If all the other providers does queue their task it will be fine.
whatever you decide

agrare:
Yes all providers use this, lets focus on the core PR and see if we can clean up how we log the targets.  I don't think we should say that the action is `"EmsRefresh completed successfully "` though since that doesn't make sense.

kbrock:
alt: https://github.com/ManageIQ/manageiq/pull/22776

PR #253, opened 2024-11-05
[WIP] Switch from secrets to rails credentials

jrafanie:
UPDATE:  The code has been updated to use fallbacks of default values in tests so CI and users don't need to actually specify `Rails.application.credentials` unless they are recording new cassettes.

TODO: Verify we can rerecord the cassettes using this PR.

Here's what you have to do:

1) To run the tests, do nothing, we have defaults specified to match the cassettes.

To record updated cassettes:
1) Apply the commit here to switch to credentials
2) Edit rails credentials in the rails application (manageiq) in the test environment:
IMPORTANT: cd to manageiq and not the rails engine directory.  It looks for the encryption key and encrypted
credentials file relative to the rails application, not the engine.

EDITOR=vi be rails credentials:edit --environment test

Specify the real values in your editor(change the values below):

```
autosde:
  appliance_host: autosde-appliance-host
  site_manager_user: autosde
  site_manager_password: change_me
```

Save this.  It should generate the following files in the rails app:

* config/credentials/test.key (if not previously created)
* an encrypted credentials file in config/credentials/test.yml.enc

Both files should be .gitignored.

You can now run record updated cassettes.


agrare:
@jrafanie yeah we need some way to have default credentials for vcrs but be able to override that in order to re-record the VCRs using real credentials.

jrafanie:
> @jrafanie yeah we need some way to have default credentials for vcrs but be able to override that in order to re-record the VCRs using real credentials.

Default credentials feels like something that's separated from rails credentials.  You could theoretically run `EDITOR=vi be rails credentials:edit --environment test` to set defaults like I did above, then ship the `config/credentials/test.key` and `config/credentials/test.yml.enc` where the `.key` is like our v2_key and the .yml.enc is the encrypted yaml of the credentials for use in CI or similar. 

There are problems with this though.  You wouldn't be able to run the tests at all without first following those steps.  I wonder if it makes more sense changing usage of `Rails.application.secrets` or `Rails.application.credential` to have a default value if `nil` in the code/test that uses it so you only need it when re-recording VCRs.

In other words:

```ruby
Rails.application.credentials.dig(:autosde, :appliance_host) || "autosde-appliance-host"
```

jrafanie:
> In other words:
> 
> ```ruby
> Rails.application.credentials.dig(:autosde, :appliance_host) || "autosde-appliance-host"
> ```

I'll just do it as additional commits... it's easier to talk about code once it's in front of you

spec/spec_helper.rb
```diff
@@ -14,9 +14,26 @@
...
-    config.define_cassette_placeholder(Rails.application.secrets.autosde_defaults[secret]) do
-      Rails.application.secrets.autosde[secret]
+  defaults = {
+    "appliance_host"        => "autosde-appliance-host",
+    "site_manager_user"     => "autosde",
+    "site_manager_password" => "change_me"
+  }
+  (Rails.application.credentials.autosde_defaults || defaults).each do |key, value|
+    config.define_cassette_placeholder(key) do
+      Rails.application.credentials.dig(:autosde, key) || value
```
jrafanie, line 24: Note, I don't know exactly what it's doing here and realize we have the methods just below... but this needs to work in two scenarios:

1) User or CI with no Rails credentials defined:
  * no config/master.key, no config/credentials.yml.enc
  * Or based on env, no config/credentals/test.key or config/credentials/test.yml.enc
2) User or CI with rails credentials defined with the same format as seen here.

jrafanie:
@agrare @nasark No rush but if one of you is able to rerecord cassettes, it would be great if you could verify the process I listed above where you run `rails credentials:edit` in the main app repo to set the correct actual credentials works when you try to update the cassettes.

spec/models/manageiq/providers/autosde/storage_manager/refresher_v2_spec.rb
```diff
@@ -4,10 +4,10 @@
   let(:ems) do
     FactoryBot.create(:autosde_storage_manager,
                       :with_autosde_credentials,
-                      :hostname => Rails.application.secrets.autosde[:appliance_host])
+                      :hostname => credentials_autosde_host)
```
jrafanie, line 7: I needed to change this and realized the test file's name was incorrect and was never run in CI or locally unless you specifically ran the path to the file.

spec/models/manageiq/providers/autosde/storage_manager/refresher_v2_spec.rb
```diff
@@ -4,10 +4,10 @@
...
   end
 
-  describe "#refresh - autosde gem v2" do
+  xcontext "#refresh - autosde gem v2 - TODO: Did this ever work? Committed with incorrect filename in 23db5ed" do
```
jrafanie, line 10: Disabled the whole test for now since it's not been run, see above.

spec/spec_helper.rb
jrafanie: In other words, if someone can run this with actual credentials to update cassettes, we can refactor it to minimize the duplication.  I didn't want to change more than I had to.

agrare: Since we won't have `Rails.application.credentials.autosde_defaults` I think we can simplify this to just `defaults.each` the rest looks right :+1:  use a credentials value if it exists otherwise use the default value.

spec/spec_helper.rb
```diff
@@ -14,9 +14,26 @@
   config.default_cassette_options = {:record => :none, :allow_unused_http_interactions => true}
   config.hook_into :webmock
 
-  Rails.application.secrets.autosde_defaults.keys.each do |secret|
-    config.define_cassette_placeholder(Rails.application.secrets.autosde_defaults[secret]) do
-      Rails.application.secrets.autosde[secret]
+  defaults = {
+    "appliance_host"        => "autosde-appliance-host",
+    "site_manager_user"     => "autosde",
+    "site_manager_password" => "change_me"
```
agrare, line 20: If you didn't want to "duplicate" the values you could always
```suggestion
    "appliance_host"        => credentials_autosde_host,
    "site_manager_user"     => credentials_autosde_user,
    "site_manager_password" => credentials_autosde_password
```

jrafanie: Yeah, I guess I wanted to see what the usage was down below to better understand if i could eliminate it entirely and do it inline.

jrafanie:
Ok, removing WIP.  I still need to record updated cassettes using this procedure but I think all of the concerns raised here have been addressed.

agrare:
@jrafanie and just to make sure I understand how the new credentials system works would we have to use `rails credentials:edit --environment test` here?  Can that be per-plugin or is it for the whole app?  We should probably include something in the plugin/providers developers guide on how to re-record vcrs

jrafanie:
> @jrafanie and just to make sure I understand how the new credentials system works would we have to use `rails credentials:edit --environment test` here? Can that be per-plugin or is it for the whole app? We should probably include something in the plugin/providers developers guide on how to re-record vcrs

rails credentials are for the rails application so I guess you'd have some .gitignore(d) credentials for each plugin.  I need to add the .gitignore for the main app.  I'll do that next.

Fryguy:
We also need to update the plugin generator.

For some reason I'm not a fan of the defaults in the code as opposed to the YAML file, but in the end it doesn't make much difference, and I won't let that hold up a merge. In the end as long as @agrare is happy then I'm happy 😅

jrafanie:
> We also need to update the plugin generator.

Yes, it's on the list.  I wanted to get a concrete example done and then it's on to all the other places we use secrets.  I'm going to try to find an environment I can try to rerecord cassettes at some point.
 
> For some reason I'm not a fan of the defaults in the code as opposed to the YAML file, but in the end it doesn't make much difference, and I won't let that hold up a merge. In the end as long as @agrare is happy then I'm happy 😅

I don't like it either.  One alternative is requiring CI/devs create their credentials for use with the test but that seems cumbersome.

Fryguy:
However I would like to have the "real" secrets gitignored ahead of time in core.

jrafanie:
> However I would like to have the "real" secrets gitignored ahead of time in core.

On the way...

jrafanie:
@agrare @Fryguy see: https://github.com/ManageIQ/manageiq/pull/23263

jrafanie:
Leaving as wip until I can run through recording cassettes along with the updated guide: https://github.com/ManageIQ/guides/pull/553

jrafanie:
Closing in favor of https://github.com/ManageIQ/manageiq/pull/23496 since Rails credentials is geared more towards teams sharing encrypted credentials, which is not what we need for VCR secrets.

PR #270, opened 2026-06-12
Add autosde storage factory methods

asirvadAbrahamVarghese:
This is to help create data for Cypress tests in https://github.com/ManageIQ/manageiq-ui-classic/pull/10095
Requires: 
- [x] https://github.com/ManageIQ/manageiq/pull/23878

@miq-bot add-label enhancement
@miq-bot add-label test

### DIFF
--- a/spec/factories/storage_resource.rb
+++ b/spec/factories/storage_resource.rb
@@ -1,0 +1,5 @@
+FactoryBot.define do
+  factory :autosde_storage_resource,
+          :parent => :storage_resource,
+          :class  => "ManageIQ::Providers::Autosde::StorageManager::StorageResource"
+end
--- a/spec/factories/storage_service.rb
+++ b/spec/factories/storage_service.rb
@@ -1,0 +1,5 @@
+FactoryBot.define do
+  factory :autosde_storage_service,
+          :parent => :storage_service,
+          :class  => "ManageIQ::Providers::Autosde::StorageManager::StorageService"
+end
--- a/spec/models/manageiq/providers/autosde/storage_service_spec.rb
+++ b/spec/models/manageiq/providers/autosde/storage_service_spec.rb
@@ -12,21 +12,20 @@ describe ManageIQ::Providers::Autosde::StorageManager::StorageService do
   end
 
   let(:test_resource1) do
-    FactoryBot.create(:storage_resource, :ext_management_system => ems, :ems_ref => "555")
+    FactoryBot.create(:autosde_storage_resource, :ext_management_system => ems, :ems_ref => "555")
   end
 
   let(:test_resource2) do
-    FactoryBot.create(:storage_resource, :ext_management_system => ems, :ems_ref => "666")
+    FactoryBot.create(:autosde_storage_resource, :ext_management_system => ems, :ems_ref => "666")
   end
 
   let(:test_resource3) do
-    FactoryBot.create(:storage_resource, :ext_management_system => ems, :ems_ref => "777")
+    FactoryBot.create(:autosde_storage_resource, :ext_management_system => ems, :ems_ref => "777")
   end
 
   before(:each) do
     @test_service = FactoryBot.create(
-      :storage_service,
-      :type                  => "ManageIQ::Providers::Autosde::StorageManager::StorageService",
+      :autosde_storage_service,
       :ext_management_system => ems,
       :storage_resources     => [test_resource1, test_resource2],
       :capabilities          => {'cap1' => "True", 'cap2' => "False"}


spec/factories/autosde_storage_resource.rb
```diff
@@ -0,0 +1,18 @@
+FactoryBot.define do
+  factory :autosde_storage_resource,
+          :class => "ManageIQ::Providers::Autosde::StorageManager::StorageResource" do
+    sequence(:name) { |n| "autosde_storage_resource_#{seq_padded_for_sorting(n)}" }
+    sequence(:ems_ref) { |n| "autosde-resource-#{n}" }
+    type { "ManageIQ::Providers::Autosde::StorageManager::StorageResource" }
```
agrare: Do you need `type {}` here and also `:class` above?

spec/factories/autosde_storage_resource.rb
```diff
@@ -0,0 +1,18 @@
...
+    logical_free  { 500_000_000_000 }
+    logical_total { 1_000_000_000_000 }
+
+    transient do
+      with_capabilities { {} }
+    end
```
agrare: Could you just do `capabilities { {} }` above?  `capabilities` is a proper column so I don't think we need transient but maybe you're doing something else

spec/factories/autosde_storage_resource.rb
```diff
@@ -0,0 +1,18 @@
+FactoryBot.define do
+  factory :autosde_storage_resource,
+          :class => "ManageIQ::Providers::Autosde::StorageManager::StorageResource" do
+    sequence(:name) { |n| "autosde_storage_resource_#{seq_padded_for_sorting(n)}" }
+    sequence(:ems_ref) { |n| "autosde-resource-#{n}" }
+    type { "ManageIQ::Providers::Autosde::StorageManager::StorageResource" }
+    logical_free  { 500_000_000_000 }
+    logical_total { 1_000_000_000_000 }
```
agrare: This seems like something which could be set by the test using this factory and doesn't need to be the same for all of them

spec/factories/autosde_storage_resource.rb
asirvadAbrahamVarghese: I had no clue. Yeah, just the class is fine, no need for the type

spec/factories/autosde_storage_resource.rb
asirvadAbrahamVarghese: I was just trying to create resources from cypress with capability values like:
```
        {
          name: STORAGE_RESOURCE_1,
          ems_id: storageManagerId,
          with_capabilities: {
            compression: ['True'],
            thin_provision: ['True'],
          },
        },
```

Looks like `capabilities: { {} }` works fine, just like you said:
```
        {
          name: STORAGE_RESOURCE_1,
          ems_id: storageManagerId,
          capabilities: {
            compression: ['True'],
            thin_provision: ['True'],
          },
        },
```

spec/factories/autosde_storage_resource.rb
asirvadAbrahamVarghese: Makes sense. I'm going to remove this since I don't think my use case actually needs this.
I tried it without it and nothing broke, so I guess I was just keeping it around because I thought some method in backend might depend on it

spec/factories/autosde_storage_resource.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :autosde_storage_resource,
+          :class => "ManageIQ::Providers::Autosde::StorageManager::StorageResource" do
+    capabilities { {} }
```
agrare: I don't think you need to duplicate the capabilities attribute here if you inherit from the parent factory in core

spec/factories/autosde_storage_service.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :autosde_storage_service,
+          :class => "ManageIQ::Providers::Autosde::StorageManager::StorageService" do
+    capabilities { {} }
```
agrare: Same

spec/factories/autosde_storage_resource.rb
```diff
@@ -0,0 +1,6 @@
...
+  end
```
agrare:
```suggestion
  factory :autosde_storage_resource,
          :class => "ManageIQ::Providers::Autosde::StorageManager::StorageResource",
          :parent => :storage_resource
```

spec/factories/autosde_storage_service.rb
```diff
@@ -0,0 +1,6 @@
...
+  end
```
agrare:
```suggestion
  factory :autosde_storage_service,
          :class => "ManageIQ::Providers::Autosde::StorageManager::StorageService",
          :parent => :storage_service
```

asirvadAbrahamVarghese: Ohh nice, this is neat, becoming a fan of Ruby now

agrare (approved):

agrare:
For the spec failures it looks like you just need to use the new autosde versions of these factories:
```
diff --git a/spec/models/manageiq/providers/autosde/storage_service_spec.rb b/spec/models/manageiq/providers/autosde/storage_service_spec.rb
index 9c98a8f..d9c0d3f 100644
--- a/spec/models/manageiq/providers/autosde/storage_service_spec.rb
+++ b/spec/models/manageiq/providers/autosde/storage_service_spec.rb
@@ -12,21 +12,20 @@ describe ManageIQ::Providers::Autosde::StorageManager::StorageService do
   end
 
   let(:test_resource1) do
-    FactoryBot.create(:storage_resource, :ext_management_system => ems, :ems_ref => "555")
+    FactoryBot.create(:autosde_storage_resource, :ext_management_system => ems, :ems_ref => "555")
   end
 
   let(:test_resource2) do
-    FactoryBot.create(:storage_resource, :ext_management_system => ems, :ems_ref => "666")
+    FactoryBot.create(:autosde_storage_resource, :ext_management_system => ems, :ems_ref => "666")
   end
 
   let(:test_resource3) do
-    FactoryBot.create(:storage_resource, :ext_management_system => ems, :ems_ref => "777")
+    FactoryBot.create(:autosde_storage_resource, :ext_management_system => ems, :ems_ref => "777")
   end
 
   before(:each) do
     @test_service = FactoryBot.create(
-      :storage_service,
-      :type                  => "ManageIQ::Providers::Autosde::StorageManager::StorageService",
+      :autosde_storage_service,
       :ext_management_system => ems,
```

spec/factories/storage_resource.rb
agrare, line 1: Actually I just noticed, can you rename this file and the other factory just be `storage_resource.rb` and `storage_service.rb`?
We usually keep the name of the factories the same as the base class, because it is pretty common to have multiple factories here.
Example: https://github.com/ManageIQ/manageiq-providers-vmware/blob/master/spec/factories/ems_cluster.rb isn't named `ems_cluster_vmware.rb`

asirvadAbrahamVarghese: Done, I went with the `autosde_` prefix since I saw `autosde_manager` under `spec/factories`

asirvadAbrahamVarghese:
> For the spec failures it looks like you just need to use the new autosde versions of these factories:
> 
> ```
> diff --git a/spec/models/manageiq/providers/autosde/storage_service_spec.rb b/spec/models/manageiq/providers/autosde/storage_service_spec.rb
> index 9c98a8f..d9c0d3f 100644
> --- a/spec/models/manageiq/providers/autosde/storage_service_spec.rb
> +++ b/spec/models/manageiq/providers/autosde/storage_service_spec.rb
> @@ -12,21 +12,20 @@ describe ManageIQ::Providers::Autosde::StorageManager::StorageService do
>    end
>  
>    let(:test_resource1) do
> -    FactoryBot.create(:storage_resource, :ext_management_system => ems, :ems_ref => "555")
> +    FactoryBot.create(:autosde_storage_resource, :ext_management_system => ems, :ems_ref => "555")
>    end
>  
>    let(:test_resource2) do
> -    FactoryBot.create(:storage_resource, :ext_management_system => ems, :ems_ref => "666")
> +    FactoryBot.create(:autosde_storage_resource, :ext_management_system => ems, :ems_ref => "666")
>    end
>  
>    let(:test_resource3) do
> -    FactoryBot.create(:storage_resource, :ext_management_system => ems, :ems_ref => "777")
> +    FactoryBot.create(:autosde_storage_resource, :ext_management_system => ems, :ems_ref => "777")
>    end
>  
>    before(:each) do
>      @test_service = FactoryBot.create(
> -      :storage_service,
> -      :type                  => "ManageIQ::Providers::Autosde::StorageManager::StorageService",
> +      :autosde_storage_service,
>        :ext_management_system => ems,
> ```

Thanks
but I don’t completely understand why adding `:parent => :storage_resource,` broke the parent factory usage: `FactoryBot.create(:storage_resource....` I assume autosde factory was just having a reference to inherit from the parent factory?

agrare:
>  but I don’t completely understand why adding :parent => :storage_resource, broke the parent factory usage: FactoryBot.create(:storage_resource.... I assume autosde factory was just having a reference to inherit from the parent factory?

I don't think it did, my guess is that it was adding the factory in core caused it to not default to the autosde subclass

spec/factories/storage_resource.rb
agrare: Ahh that makes sense, that was also incorrect :laughing: but no way for you to have known that

agrare (approved):

